PR #20, opened 2020-05-07
Migrate to non-blocking architecture using Spring WebFlux

ChamseddineBhd:


src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -175,17 +167,14 @@ public String getFilename() {
...
-        } catch (HttpStatusCodeException e) {
-            throw new StudyException("importCase " + e.getStatusCode() + " : " + e.getResponseBodyAsString(), e);
-        }
+        String uuid = webClient.post()
+                .uri(caseServerBaseUri + "/" + CASE_API_VERSION + "/cases/private")
+                .header(HttpHeaders.CONTENT_TYPE, MediaType.MULTIPART_FORM_DATA.toString())
+                .body(BodyInserters.fromValue(map))
+                .retrieve()
+                .bodyToMono(String.class)
+                .block();
```
geofjamg, line 176: This is forbidden!

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -280,12 +269,13 @@ boolean caseExists(UUID caseUuid) {
...
 
-        ResponseEntity<Boolean> responseEntity = caseServerRest.exchange(path,
-                HttpMethod.GET,
-                HttpEntity.EMPTY,
-                Boolean.class);
+        String caseExists = webClient.get()
+                .uri(caseServerBaseUri + path)
+                .retrieve()
+                .bodyToMono(String.class)
+                .block();
```
geofjamg, line 276: This is forbidden!

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -103,12 +115,11 @@ private String getCaseFormat(UUID caseUuid) {
...
-                HttpMethod.GET,
-                HttpEntity.EMPTY,
-                String.class);
-
-        return responseEntity.getBody();
+        return webClient.get()
+                .uri(caseServerBaseUri + path)
+                .retrieve()
+                .bodyToMono(String.class)
+                .block();
```
geofjamg, line 122: This is forbidden!

src/main/java/org/gridsuite/study/server/StudyService.java
ChamseddineBhd: I'm migarting to webFlux smoothly, so the first step is to migrate all restTemplate http requests to webClient which is non blocking client 
the .block is temporary and will be obviously removed in the next steps.

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,142 @@
...
 
     @GetMapping(value = "/studies/{studyName}/network/voltage-levels")
     @ApiOperation(value = "get the voltage levels for a given network")
     @ApiResponse(code = 200, message = "The voltage level list of the network")
-    public ResponseEntity<List<VoltageLevelAttributes>> getNetworkVoltyutageLevels(@PathVariable("studyName") String studyName) {
-        UUID networkUuid = studyService.getStudyUuid(studyName);
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studyService.getNetworkVoltageLevels(networkUuid));
+    public ResponseEntity<List<VoltageLevelAttributes>> getNetworkVoltageLevels(@PathVariable("studyName") String studyName) {
+        Mono<UUID> networkUuid = studyService.getStudyUuid(studyName);
+        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studyService.getNetworkVoltageLevels(networkUuid.block()));
```
geofjamg, line 158: You needed to block here?

ChamseddineBhd: No, that will be removed in the next commit.

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -54,20 +57,23 @@
...
+
+        return Mono.zip(studyExists, caseExists)
+                .flatMap(t -> {
+                    if (t.getT1().equals(Boolean.TRUE)) {
+                        return Mono.just(ResponseEntity.status(HttpStatus.CONFLICT).body(STUDY_ALREADY_EXISTS));
+                    } else if (t.getT2().equals(Boolean.FALSE)) {
+                        return Mono.just(ResponseEntity.status(HttpStatus.NOT_FOUND).body(CASE_DOESNT_EXISTS));
+                    } else {
+                        return studyService.createStudy(studyName, caseUuid, description)
+                                .flatMap(s -> Mono.just(ResponseEntity.ok().build()));
```
jonenst, line 74: using ```flatmap( ... -> Mono.just(...) )``` is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -54,20 +57,23 @@
...
-
-        studyService.createStudy(studyName, caseUuid, description);
-        return ResponseEntity.ok().build();
+    public Mono<ResponseEntity<Object>> createStudyFromExistingCase(@PathVariable("studyName") String studyName,
+                                                    @PathVariable("caseUuid") UUID caseUuid,
+                                                    @RequestParam("description") String description) {
+        Mono<Boolean> studyExists = studyService.studyExists(studyName);
+        Mono<Boolean> caseExists = studyService.caseExists(caseUuid);
+
+        return Mono.zip(studyExists, caseExists)
```
jonenst, line 66: did you choose on purpose to do the studyExists and caseExists calls in parallel ? Before that it was sequential.

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -54,20 +57,23 @@
...
-            throw new ResponseStatusException(HttpStatus.CONFLICT, STUDY_ALREADY_EXISTS);
-        }
-
-        if (!studyService.caseExists(caseUuid)) {
-            throw new ResponseStatusException(HttpStatus.NOT_FOUND, CASE_DOESNT_EXISTS);
-        }
-
-        studyService.createStudy(studyName, caseUuid, description);
-        return ResponseEntity.ok().build();
+    public Mono<ResponseEntity<Object>> createStudyFromExistingCase(@PathVariable("studyName") String studyName,
```
jonenst, line 60: Why change from Void to Object ?

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -54,20 +57,23 @@
...
+                .flatMap(t -> {
+                    if (t.getT1().equals(Boolean.TRUE)) {
+                        return Mono.just(ResponseEntity.status(HttpStatus.CONFLICT).body(STUDY_ALREADY_EXISTS));
+                    } else if (t.getT2().equals(Boolean.FALSE)) {
+                        return Mono.just(ResponseEntity.status(HttpStatus.NOT_FOUND).body(CASE_DOESNT_EXISTS));
+                    } else {
+                        return studyService.createStudy(studyName, caseUuid, description)
+                                .flatMap(s -> Mono.just(ResponseEntity.ok().build()));
+                    }
+                });
```
jonenst, line 76: Maybe it's easier to read if you keep the exceptions instead of returning ResponseEntities and you attach the exceptions to their respective monos (untested)

sequential
```java
studyService.studyExists(studyName).flatmap( exists -> exists ? Mono.error(...) : Mono.empty() )
.then(
  studyService.caseExists(caseUuid).flatmap( exists -> exists ? Mono.empty() : Mono.error(...) )
).then(
studyService.createStudy().thenReturn(ResponseEntity.ok().build())
)
```

I think you get parallel execution by replacing the first ```then``` with ```zipWith```:
```java
studyService.studyExists(studyName).flatmap(...)
.zipWith(
 studyService.caseExists(caseUuid).flatmap(...)
).then(
 studyService.createStudy().thenReturn(...)
)
```
```  .flatmap( exists -> exists ? Mono.error(...) : Mono.empty() )``` can be replaced by ```filter(...).switchOnEmpty(Mono.error())```, not sure which is is cleaner...

File: src/main/java/org/gridsuite/study/server/StudyController.java [hunk 9]
```diff
@@ -76,140 +82,144 @@
...
-
-        studyService.createStudy(studyName, caseFile, description);
-        return ResponseEntity.ok().build();
+    public Mono<ResponseEntity<Object>> createStudy(@PathVariable("studyName") String studyName,
+                                    @RequestParam("caseFile") MultipartFile caseFile,
+                                    @RequestParam("description") String description) throws IOException {
+        Mono<Boolean> studyExists = studyService.studyExists(studyName);
+
+        return studyExists.flatMap(b -> {
+            if (b.equals(Boolean.TRUE)) {
```
jonenst, line 91: extract to a method that asserts that the study doesn't exist: return a mono that errors when it exists, or empty when it doesn't exist ?

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
+                return Mono.just(ResponseEntity.status(HttpStatus.CONFLICT).body(STUDY_ALREADY_EXISTS));
+            }
+            try {
```
jonenst, line 94: why do we need this try catch block ?

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
             @ApiResponse(code = 404, message = "The study doesn't exist")})
-    public ResponseEntity<Study> getStudy(@PathVariable("studyName") String studyName) {
-        Study study = studyService.getStudy(studyName);
-        if (study == null) {
-            return ResponseEntity.notFound().build();
-        }
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(study);
+    public Mono<ResponseEntity<Study>> getStudy(@PathVariable("studyName") String studyName) {
+        Mono<Study> studyMono = studyService.getStudy(studyName);
+        return studyMono.flatMap(s -> Mono.just(ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(s)))
```
jonenst, line 110: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
             @ApiParam(value = "centerLabel") @RequestParam(name = "centerLabel", defaultValue = "false") boolean centerLabel,
             @ApiParam(value = "diagonalLabel") @RequestParam(name = "diagonalLabel", defaultValue = "false") boolean diagonalLabel,
             @ApiParam(value = "topologicalColoring") @RequestParam(name = "topologicalColoring", defaultValue = "false") boolean topologicalColoring) {
-        UUID networkUuid = studyService.getStudyUuid(studyName);
-
-        byte[] svg = studyService.getVoltageLevelSvg(networkUuid, voltageLevelId, useName, centerLabel, diagonalLabel, topologicalColoring);
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_XML).body(svg);
+        return studyService.getStudyUuid(studyName)
+                .flatMap(uuid -> studyService.getVoltageLevelSvg(uuid, voltageLevelId, useName, centerLabel, diagonalLabel, topologicalColoring))
+                .flatMap(svg -> Mono.just(ResponseEntity.ok().contentType(MediaType.APPLICATION_XML).body(svg)))
```
jonenst, line 133: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
             @ApiParam(value = "centerLabel") @RequestParam(name = "centerLabel", defaultValue = "false") boolean centerLabel,
             @ApiParam(value = "diagonalLabel") @RequestParam(name = "diagonalLabel", defaultValue = "false") boolean diagonalLabel,
             @ApiParam(value = "topologicalColoring") @RequestParam(name = "topologicalColoring", defaultValue = "false") boolean topologicalColoring) {
-        UUID networkUuid = studyService.getStudyUuid(studyName);
-        String svgAndMetadata = studyService.getVoltageLevelSvgAndMetadata(networkUuid, voltageLevelId, useName, centerLabel, diagonalLabel, topologicalColoring);
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(svgAndMetadata);
+        return studyService.getStudyUuid(studyName)
+                .flatMap(uuid -> studyService.getVoltageLevelSvgAndMetadata(uuid, voltageLevelId, useName, centerLabel, diagonalLabel, topologicalColoring))
+                .flatMap(svgAndMetadata -> Mono.just(ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(svgAndMetadata)))
```
jonenst, line 149: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
     @ApiOperation(value = "get the voltage levels for a given network")
     @ApiResponse(code = 200, message = "The voltage level list of the network")
-    public ResponseEntity<List<VoltageLevelAttributes>> getNetworkVoltyutageLevels(@PathVariable("studyName") String studyName) {
-        UUID networkUuid = studyService.getStudyUuid(studyName);
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studyService.getNetworkVoltageLevels(networkUuid));
-
+    public Mono<ResponseEntity<List<VoltageLevelAttributes>>> getNetworkVoltageLevels(@PathVariable("studyName") String studyName) {
+        Mono<UUID> networkUuid = studyService.getStudyUuid(studyName);
+        return networkUuid.flatMap(uuid -> {
+            return Mono.just(studyService.getNetworkVoltageLevels(uuid));
```
jonenst, line 159: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
+        }).flatMap(vls -> Mono.just(ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(vls)));
```
jonenst, line 160: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
     @ApiOperation(value = "Get Network lines graphics", produces = "application/json")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The list of lines graphics")})
-    public ResponseEntity<String> getLinesGraphics(@PathVariable("studyName") String studyName) {
-        UUID networkUuid = studyService.getStudyUuid(studyName);
-        String lineGraphics = studyService.getLinesGraphics(networkUuid);
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(lineGraphics);
+    public Mono<ResponseEntity<String>> getLinesGraphics(@PathVariable("studyName") String studyName) {
+        return studyService.getStudyUuid(studyName)
+                .flatMap(studyService::getSubstationsGraphics)
+                .flatMap(lineGraphics -> Mono.just(ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(lineGraphics)));
```
jonenst, line 169: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
     @ApiOperation(value = "Get Network substations graphics", produces = "application/json")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The list of substations graphics")})
-    public ResponseEntity<String> getSubstationsGraphic(@PathVariable("studyName") String studyName) {
-        UUID networkUuid = studyService.getStudyUuid(studyName);
-        String substationGraphics = studyService.getSubstationsGraphics(networkUuid);
-        return  ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(substationGraphics);
+    public Mono<ResponseEntity<String>> getSubstationsGraphic(@PathVariable("studyName") String studyName) {
+        return studyService.getStudyUuid(studyName)
+                .flatMap(studyService::getSubstationsGraphics)
+                .flatMap(substationGraphics -> Mono.just(ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(substationGraphics)));
```
jonenst, line 178: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
     @ApiOperation(value = "Get Network lines description", produces = "application/json")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The list of lines graphics")})
-    public ResponseEntity<String> getLinesMapData(@PathVariable("studyName") String studyName) {
-        UUID networkUuid = studyService.getStudyUuid(studyName);
-        String linesMapData = studyService.getLinesMapData(networkUuid);
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(linesMapData);
+    public Mono<ResponseEntity<String>> getLinesMapData(@PathVariable("studyName") String studyName) {
+        return studyService.getStudyUuid(studyName)
+                .flatMap(studyService::getLinesMapData)
+                .flatMap(linesMapData -> Mono.just(ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(linesMapData)));
```
jonenst, line 187: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
     @ApiOperation(value = "Get Network substations description", produces = "application/json")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The list of substations graphics")})
-    public ResponseEntity<String> getSubstationsMapData(@PathVariable("studyName") String studyName) {
-        UUID networkUuid = studyService.getStudyUuid(studyName);
-        String substationMapData = studyService.getSubstationsMapData(networkUuid);
-        return  ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(substationMapData);
+    public Mono<ResponseEntity<String>> getSubstationsMapData(@PathVariable("studyName") String studyName) {
+        return studyService.getStudyUuid(studyName)
+                .flatMap(studyService::getSubstationsMapData)
+                .flatMap(substationMapData -> Mono.just(ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(substationMapData)));
```
jonenst, line 196: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The switch is updated")})
-    public ResponseEntity<Void> changeSwitchState(@PathVariable("studyName") String studyName,
-                                                  @PathVariable("switchId") String switchId,
-                                                  @RequestParam("open") boolean open) {
-        studyService.changeSwitchState(studyName, switchId, open);
-        return  ResponseEntity.ok().build();
+    public Mono<ResponseEntity<Void>> changeSwitchState(@PathVariable("studyName") String studyName,
+                                                          @PathVariable("switchId") String switchId,
+                                                          @RequestParam("open") boolean open) {
+        return studyService.changeSwitchState(studyName, switchId, open).flatMap(e -> Mono.just(ResponseEntity.ok().build()));
```
jonenst, line 205: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
     }
 
     @PutMapping(value = "/studies/{studyName}/loadflow/run")
     @ApiOperation(value = "run loadflow on study", produces = "application/json")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The loadflow has started")})
-    public ResponseEntity<Void> runLoadFlow(@PathVariable("studyName") String studyName) {
-        studyService.runLoadFlow(studyName);
-        return  ResponseEntity.ok().build();
+    public Mono<ResponseEntity<Void>> runLoadFlow(@PathVariable("studyName") String studyName) {
+        return studyService.runLoadFlow(studyName).flatMap(e -> Mono.just(ResponseEntity.ok().build()));
```
jonenst, line 212: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +82,144 @@
...
     @ApiOperation(value = "Update the study name", produces = "application/json")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The updated study")})
-    public ResponseEntity<Study> renameStudy(@PathVariable("studyName") String studyName,
-                                                        @RequestBody RenameStudyAttributes renameStudyAttributes) {
-        Study study = studyService.renameStudy(studyName, renameStudyAttributes.getNewStudyName());
-        return  ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(study);
+    public Mono<ResponseEntity<Study>> renameStudy(@PathVariable("studyName") String studyName,
+                                                   @RequestBody RenameStudyAttributes renameStudyAttributes) {
+        Mono<Study> studyMono = studyService.renameStudy(studyName, renameStudyAttributes.getNewStudyName());
+        return studyMono.flatMap(study -> Mono.just(ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(study)))
```
jonenst, line 221: flatmap( ... -> Mono.just(...) ) is an antipattern

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -67,106 +64,88 @@ public StudyService(
...
     }
 
-    void createStudy(String studyName, UUID caseUuid, String description) {
-        NetworkInfos networkInfos = persistentStore(caseUuid);
-        String caseFormat = getCaseFormat(caseUuid);
-        Study study = new Study(studyName, networkInfos.getNetworkUuid(), networkInfos.getNetworkId(), description, caseFormat, caseUuid, false);
-        studyRepository.insert(study);
+    Mono<Study> createStudy(String studyName, UUID caseUuid, String description) {
+        Mono<NetworkInfos> networkInfos = persistentStore(caseUuid);
+        Mono<String> caseFormat = getCaseFormat(caseUuid);
```
jonenst, line 93: maybe cleaner with zipWith ?

src/main/java/org/gridsuite/study/server/StudyController.java
ChamseddineBhd: Mono.error() --> response status 500 INTERNAL EXCEPTION
we want to send the suitable response status 409 conflict or NotFound 404 to know what is the real problem.
ok with the proposition zipWith it's better for a sequential check.

jonenst:
I think that mono.error() is handled like throwing exceptions: it looks for
a handler to map the exception to a http status no?

Le mar. 26 mai 2020 à 21:17, ChamseddineBhd <notifications@github.com> a
écrit :

src/main/java/org/gridsuite/study/server/StudyController.java
jonenst: or something like
```
Mono.when(assertStudyNotExists, assertCaseExists)
    .then( studyService.createStudy() ) ;
```
?

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -38,14 +36,17 @@
...
+    public StudyController(StudyService studyService) {
+        this.studyService = studyService;
+    }
 
     @GetMapping(value = "/studies")
     @ApiOperation(value = "Get all studies")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The list of studies")})
-    public ResponseEntity<List<StudyInfos>> getStudyList() {
-        List<StudyInfos> studies = studyService.getStudyList();
+    public ResponseEntity<Flux<StudyInfos>> getStudyList() {
```
jonenst, line 48: What does this do ? instead of ```ResponseEntity<Flux<StudyInfos>>``` I would have expected the Flux to be the outer type

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +68,132 @@
...
-            throw new ResponseStatusException(HttpStatus.CONFLICT, STUDY_ALREADY_EXISTS);
-        }
-
-        studyService.createStudy(studyName, caseFile, description);
-        return ResponseEntity.ok().build();
+    public Mono<ResponseEntity<Void>> createStudy(@PathVariable("studyName") String studyName,
+                                    @RequestParam("caseFile") MultipartFile caseFile,
+                                    @RequestParam("description") String description) {
+        return studyService.assertStudyNotExists(studyName)
+                .flatMap(e -> studyService.createStudy(studyName, caseFile, description).map(s -> ResponseEntity.ok().build()));
```
jonenst, line 75: use .then instead of flatmap

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -54,20 +55,11 @@
...
-            throw new ResponseStatusException(HttpStatus.NOT_FOUND, CASE_DOESNT_EXISTS);
-        }
-
-        studyService.createStudy(studyName, caseUuid, description);
-        return ResponseEntity.ok().build();
+    public Mono<ResponseEntity<Void>> createStudyFromExistingCase(@PathVariable("studyName") String studyName,
+                                                    @PathVariable("caseUuid") UUID caseUuid,
+                                                    @RequestParam("description") String description) {
+        return Mono.when(studyService.assertStudyNotExists(studyName), studyService.assertCaseExists(caseUuid))
+                .then(studyService.createStudy(studyName, caseUuid, description).map(s -> ResponseEntity.ok().build()));
```
jonenst, line 62: use thenReturn instead of map

src/main/java/org/gridsuite/study/server/StudyController.java
jonenst: use thenReturn instead of map

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +68,132 @@
...
-    public ResponseEntity<Study> getStudy(@PathVariable("studyName") String studyName) {
-        Study study = studyService.getStudy(studyName);
-        if (study == null) {
-            return ResponseEntity.notFound().build();
-        }
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(study);
+    public Mono<ResponseEntity<Study>> getStudy(@PathVariable("studyName") String studyName) {
+        Mono<Study> studyMono = studyService.getStudy(studyName);
+        return studyMono.map(s -> ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(s))
+                .switchIfEmpty(Mono.error(new ResponseStatusException(HttpStatus.NOT_FOUND)));
```
jonenst, line 86: put the switchIfEmpty before the map

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +68,132 @@
...
     }
 
     @DeleteMapping(value = "/studies/{studyName}")
     @ApiOperation(value = "delete the study")
     @ApiResponse(code = 200, message = "Study deleted")
-    public ResponseEntity<Study> deleteStudy(@PathVariable("studyName") String studyName) {
-        studyService.deleteStudy(studyName);
-        return ResponseEntity.ok().build();
+    public Mono<ResponseEntity<Void>> deleteStudy(@PathVariable("studyName") String studyName) {
+        return studyService.deleteStudy(studyName).map(e -> ResponseEntity.ok().build());
```
jonenst, line 93: thenReturn

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +68,132 @@
...
             @ApiParam(value = "diagonalLabel") @RequestParam(name = "diagonalLabel", defaultValue = "false") boolean diagonalLabel,
             @ApiParam(value = "topologicalColoring") @RequestParam(name = "topologicalColoring", defaultValue = "false") boolean topologicalColoring) {
-        UUID networkUuid = studyService.getStudyUuid(studyName);
-
-        byte[] svg = studyService.getVoltageLevelSvg(networkUuid, voltageLevelId, useName, centerLabel, diagonalLabel, topologicalColoring);
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_XML).body(svg);
+        return studyService.getStudyUuid(studyName)
+                .flatMap(uuid -> studyService.getVoltageLevelSvg(uuid, voltageLevelId, useName, centerLabel, diagonalLabel, topologicalColoring))
+                .map(svg -> ResponseEntity.ok().contentType(MediaType.APPLICATION_XML).body(svg))
+                .onErrorReturn(ResponseEntity.notFound().build());
```
jonenst, line 109: why return 404 for all errors ?

src/main/java/org/gridsuite/study/server/StudyController.java
jonenst: according to https://docs.spring.io/spring/docs/current/spring-framework-reference/web-reactive.html#webflux-ann-responseentity

> WebFlux supports using a single value reactive type to produce the ResponseEntity asynchronously, and/or single and multi-value reactive types for the body.

So it should be Mono<ResponseEntity<Flux<StudyInfos>>

jonenst: Maybe we can remove (separate PR?) all the ResponseEntity to simplify things ? I don't think we need it because @ResponseBody should put the correct JSON content-type headers, and we can use exceptions for http statuses. Then we can only return Flux<StudyInfo> or Mono<Study> or other direct types

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +68,132 @@
...
             @ApiParam(value = "centerLabel") @RequestParam(name = "centerLabel", defaultValue = "false") boolean centerLabel,
             @ApiParam(value = "diagonalLabel") @RequestParam(name = "diagonalLabel", defaultValue = "false") boolean diagonalLabel,
             @ApiParam(value = "topologicalColoring") @RequestParam(name = "topologicalColoring", defaultValue = "false") boolean topologicalColoring) {
-        UUID networkUuid = studyService.getStudyUuid(studyName);
-        String svgAndMetadata = studyService.getVoltageLevelSvgAndMetadata(networkUuid, voltageLevelId, useName, centerLabel, diagonalLabel, topologicalColoring);
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(svgAndMetadata);
+        return studyService.getStudyUuid(studyName)
+                .flatMap(uuid -> studyService.getVoltageLevelSvgAndMetadata(uuid, voltageLevelId, useName, centerLabel, diagonalLabel, topologicalColoring))
+                .map(svgAndMetadata -> ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(svgAndMetadata))
+                .onErrorReturn(ResponseEntity.notFound().build());
```
jonenst, line 125: why return 404 for all errors ?

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +68,132 @@
...
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The switch is updated")})
-    public ResponseEntity<Void> changeSwitchState(@PathVariable("studyName") String studyName,
-                                                  @PathVariable("switchId") String switchId,
-                                                  @RequestParam("open") boolean open) {
-        studyService.changeSwitchState(studyName, switchId, open);
-        return  ResponseEntity.ok().build();
+    public Mono<ResponseEntity<Void>> changeSwitchState(@PathVariable("studyName") String studyName,
+                                                          @PathVariable("switchId") String switchId,
+                                                          @RequestParam("open") boolean open) {
+        return studyService.changeSwitchState(studyName, switchId, open).map(e -> ResponseEntity.ok().build());
```
jonenst, line 179: thenReturn

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -76,140 +68,132 @@
...
     }
 
     @PutMapping(value = "/studies/{studyName}/loadflow/run")
     @ApiOperation(value = "run loadflow on study", produces = "application/json")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The loadflow has started")})
-    public ResponseEntity<Void> runLoadFlow(@PathVariable("studyName") String studyName) {
-        studyService.runLoadFlow(studyName);
-        return  ResponseEntity.ok().build();
+    public Mono<ResponseEntity<Void>> runLoadFlow(@PathVariable("studyName") String studyName) {
+        return studyService.runLoadFlow(studyName).map(e -> ResponseEntity.ok().build());
```
jonenst, line 186: thenReturn

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -67,106 +65,93 @@ public StudyService(
...
+        this.caseServerBaseUri = caseServerBaseUri;
+        this.singleLineDiagramServerBaseUri = singleLineDiagramServerBaseUri;
+        this.networkConversionServerBaseUri = networkConversionServerBaseUri;
+        this.geoDataServerBaseUri = geoDataServerBaseUri;
+        this.networkMapServerBaseUri = networkMapServerBaseUri;
+        this.networkModificationServerBaseUri = networkModificationServerBaseUri;
+        this.loadFlowServerBaseUri = loadFlowServerBaseUri;
+
+        ExchangeStrategies exchangeStrategies = ExchangeStrategies.builder()
+                .codecs(configurer -> configurer.defaultCodecs().maxInMemorySize(1024 * 1024 * 10)).build();
```
jonenst, line 80: Why do we need ```maxInMemorySize(1024 * 1024 * 10)```

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -67,106 +65,93 @@ public StudyService(
...
-
-        networkModificationServerRest = restTemplateBuilder.build();
-        networkModificationServerRest.setUriTemplateHandler(new DefaultUriBuilderFactory(networkModificationServerBaseUri));
-
-        loadFlowServerRest = restTemplateBuilder.build();
-        loadFlowServerRest.setUriTemplateHandler(new DefaultUriBuilderFactory(loadFlowServerBaseUri));
+            @Value("${backing-services.loadflow.base-uri:http://loadflow-server/}") String loadFlowServerBaseUri,
+            NetworkStoreService networkStoreClient,
+            StudyRepository studyRepository) {
+        this.caseServerBaseUri = caseServerBaseUri;
```
jonenst, line 71: Maybe better to use @Autowired on the fields instead if we only assign them ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -44,20 +43,19 @@
...
-    private RestTemplate loadFlowServerRest;
+    String caseServerBaseUri;
+    String singleLineDiagramServerBaseUri;
+    String networkConversionServerBaseUri;
+    String geoDataServerBaseUri;
+    String networkMapServerBaseUri;
+    String networkModificationServerBaseUri;
+    String loadFlowServerBaseUri;
 
-    @Autowired
```
jonenst, line 56: Why move this to the constructor ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -67,106 +65,93 @@ public StudyService(
...
-    void createStudy(String studyName, MultipartFile caseFile, String description) throws IOException {
-        UUID caseUUid = importCase(caseFile);
-        NetworkInfos networkInfos = persistentStore(caseUUid);
-        String caseFormat = getCaseFormat(caseUUid);
-        Study study = new Study(studyName, networkInfos.getNetworkUuid(), networkInfos.getNetworkId(), description, caseFormat, caseUUid, true);
-        studyRepository.insert(study);
+    Mono<Study> createStudy(String studyName, MultipartFile caseFile, String description) {
+        Mono<UUID> caseUUid;
+        try {
+            caseUUid = importCase(caseFile);
```
jonenst, line 117: Can we avoid try catch and instead rely on flux/mono errors ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -67,106 +65,93 @@ public StudyService(
...
+        } catch (IOException e) {
+            return Mono.error(new StudyException("error when importing the case"));
+        }
+        return caseUUid.flatMap(uuid -> {
+            Mono<NetworkInfos> networkInfos = persistentStore(uuid);
+            Mono<String> caseFormat = getCaseFormat(uuid);
+            return Mono.zip(networkInfos, caseFormat)
+                    .flatMap(t -> {
```
jonenst, line 125: reduce nesting by moving the inner flatmap below the outer flatmap

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -67,106 +65,93 @@ public StudyService(
...
-                return responseEntity.getBody();
-            } else {
-                return Collections.emptyList();
+        return studyMono.flatMap(study -> {
+            String path = UriComponentsBuilder.fromPath(DELIMITER + CASE_API_VERSION + "/cases/{caseUuid}")
+                    .buildAndExpand(study.getCaseUuid())
+                    .toUriString();
+
+            if (study.isCasePrivate()) {
+                return webClient.delete()
```
jonenst, line 145: If i understand correctly, when the case is private we don't delete the study ? This is a bug, and there should be a test to show the bug

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -67,106 +65,93 @@ public StudyService(
...
-        }
+            return studyRepository.deleteByName(studyName);
+        });
     }
 
-    UUID importCase(MultipartFile multipartFile) throws IOException {
-        HttpHeaders requestHeaders = new HttpHeaders();
-        requestHeaders.setContentType(MediaType.MULTIPART_FORM_DATA);
-
+    Mono<UUID> importCase(MultipartFile multipartFile) throws IOException {
```
jonenst, line 154: An improvement could be to use the reactive APIs to get the file so that starting the call to the case server doesn't have to wait until the whole file has been received on the study server

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -180,64 +165,56 @@ public String getFilename() {
...
-                HttpEntity.EMPTY,
-                NetworkInfos.class);
-        return responseEntity.getBody();
+        return webClient.post()
+                .uri(networkConversionServerBaseUri + path)
+                .retrieve()
+                .bodyToMono(NetworkInfos.class);
     }
 
     List<VoltageLevelAttributes> getNetworkVoltageLevels(UUID networkUuid) {
```
jonenst, line 220: should we convert this to a nonblocking call ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -252,147 +229,150 @@ private NetworkInfos persistentStore(UUID caseUuid) {
...
     }
 
-    void runLoadFlow(String studyName) {
-        UUID networkUuid = getStudyUuid(studyName);
+    @Transactional
+    public Mono<Study> renameStudy(String studyName, String newStudyName) {
+        Mono<Study> studyMono = studyRepository.findByName(studyName);
+        return studyMono.flatMap(study -> {
+            study.setName(newStudyName);
+            studyRepository.deleteByName(studyName).subscribe();
```
jonenst, line 324: why subscribe manually ? Instead chain the monos

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -252,147 +229,150 @@ private NetworkInfos persistentStore(UUID caseUuid) {
...
+            return Mono.just(study);
+        }).flatMap(studyRepository::insert);
+    }
 
-        String path = UriComponentsBuilder.fromPath("/" + LOADFLOW_API_VERSION + "/networks/{networkUuid}/run")
-                .buildAndExpand(networkUuid)
-                .toUriString();
+    Mono<UUID> getStudyUuid(String studyName) {
+        Mono<Study> studyMono = studyRepository.findByName(studyName);
+        return studyMono.flatMap(study -> Mono.just(study.getNetworkUuid()))
```
jonenst, line 331: flatmap Mono.just

src/main/java/org/gridsuite/study/server/StudyService.java
jonenst: Also why remove the STUDY_DOESNT_EXISTS error ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -252,147 +229,150 @@ private NetworkInfos persistentStore(UUID caseUuid) {
...
-            throw new StudyException(STUDY_DOESNT_EXISTS);
-        }
-        Study study = studyOpt.get();
-        study.setName(newStudyName);
-        studyRepository.insert(study);
-        studyRepository.deleteByName(studyName);
-        return study;
+    Mono<Boolean> studyExists(String studyName) {
+        return getStudy(studyName).map(s -> true)
+                .switchIfEmpty(Mono.just(false));
```
jonenst, line 338: use Mono.hasElement()  instead

src/main/java/org/gridsuite/study/server/StudySwaggerConfig.java
```diff
@@ -9,21 +9,19 @@
...
 import springfox.documentation.spring.web.plugins.Docket;
-import springfox.documentation.swagger2.annotations.EnableSwagger2;
 
 /**
  * @author Abdelsalem Hedhili <abdelsalem.hedhili at rte-france.com>
  */
 
-@Configuration
-@EnableSwagger2
+//@Configuration
```
jonenst, line 23: Add a comment saying why this is commented and how to reenable

src/main/java/org/gridsuite/study/server/StudyController.java
ChamseddineBhd: > Maybe we can remove (separate PR?) all the ResponseEntity to simplify things ? I don't think we need it because @responsebody should put the correct JSON content-type headers, and we can use exceptions for http statuses. Then we can only return Flux or Mono or other direct types

ok i'll do that in a separate PR.

src/main/java/org/gridsuite/study/server/StudyService.java
ChamseddineBhd: If not the buffering limits on the webflux codecs will be exceeded.

src/main/java/org/gridsuite/study/server/StudyService.java
ChamseddineBhd: Field injection is not recommended.

src/main/java/org/gridsuite/study/server/StudyService.java
ChamseddineBhd: that cannot be possible because the inner flatMap depends on uuid of the outer flatMap.

src/main/java/org/gridsuite/study/server/StudyService.java
ChamseddineBhd: if the case is public we don't delete it *  but we delete the study.

src/main/java/org/gridsuite/study/server/StudyService.java
ChamseddineBhd: of course !

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -38,36 +36,30 @@
...
 
     @GetMapping(value = "/studies")
     @ApiOperation(value = "Get all studies")
     @ApiResponses(value = {@ApiResponse(code = 200, message = "The list of studies")})
-    public ResponseEntity<List<StudyInfos>> getStudyList() {
-        List<StudyInfos> studies = studyService.getStudyList();
-        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studies);
+    public Mono<ResponseEntity<Flux<StudyInfos>>> getStudyList() {
+        Flux<StudyInfos> studies = studyService.getStudyList();
+        return Mono.just(ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studies));
```
jonenst, line 50: Use Flux.collect or a variant to make a chain from start to the end

jonenst: You can change the return type from Mono<ResponseEntity<Flux<StudyInfos>>>> to Mono<ResponseEntity<List<StudyInfos>>>>

src/main/java/org/gridsuite/study/server/StudyService.java
jonenst: What are the limits? What happens when they are exceeded ? In what case do we exceed them ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -44,355 +42,330 @@
...
+        return studyMono.flatMap(study -> {
+            String path = UriComponentsBuilder.fromPath(DELIMITER + CASE_API_VERSION + "/cases/{caseUuid}")
+                    .buildAndExpand(study.getCaseUuid())
+                    .toUriString();
+            if (study.isCasePrivate()) {
+                return webClient.delete()
+                        .uri(caseServerBaseUri + path)
+                        .retrieve()
+                        .bodyToMono(Void.class)
+                        .then(studyRepository.deleteByName(studyName));
```
jonenst, line 142: in this case the studyRepository.deleteByName is called twice, it's a bug

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -44,355 +42,330 @@
...
-
-            if (responseEntity.getStatusCode() == HttpStatus.OK) {
-                return responseEntity.getBody();
-            } else {
-                return Collections.emptyList();
+    Mono<Void> deleteStudy(String studyName) {
+        Mono<Study> studyMono = studyRepository.findByName(studyName);
+
+        return studyMono.flatMap(study -> {
+            String path = UriComponentsBuilder.fromPath(DELIMITER + CASE_API_VERSION + "/cases/{caseUuid}")
```
jonenst, line 134: compute path only if isCasePrivate()

src/test/java/org/gridsuite/study/server/StudyTest.java
```diff
@@ -101,225 +86,222 @@
...
-            MockMultipartFile mockFile = new MockMultipartFile("caseFile", TEST_FILE, "text/plain", inputStream);
-            MvcResult mvcResult = mvc.perform(MockMvcRequestBuilders.multipart(STUDIES_URL, "s2")
-                    .file(mockFile)
-                    .param(DESCRIPTION, "desc"))
-                    .andExpect(status().isOk())
-                    .andReturn();
-        }
-
-        //Import the same case -> 409 conflict
-        try (InputStream inputStream = getClass().getResourceAsStream("/src/test/resources/testCase.xiidm")) {
```
jonenst, line 267: restore this test

src/main/java/org/gridsuite/study/server/RestResponseEntityExceptionHandler.java
```diff
@@ -1,25 +1,28 @@
...
 import org.springframework.web.bind.annotation.ExceptionHandler;
-import org.springframework.web.context.request.WebRequest;
-import org.springframework.web.servlet.mvc.method.annotation.ResponseEntityExceptionHandler;
-
-import static org.gridsuite.study.server.StudyConstants.STUDY_DOESNT_EXISTS;
+import org.springframework.web.server.ServerWebExchange;
 
 @ControllerAdvice
-public class RestResponseEntityExceptionHandler extends ResponseEntityExceptionHandler {
+@RequiredArgsConstructor
```
jonenst, line 13: is this really needed ?

ChamseddineBhd: No

PR #34, opened 2020-07-21
link the user to the created study if private

AbdelHedhili:
Signed-off-by: AbdelHedhili <abdelsalem.hedhili@rte-france.com>


AbdelHedhili:
I store the userName so we can print it in in the study informations or things like that. For exemple if an admin can see all studies (even the private ones) it'll be nice to see the name of the owner.

geofjamg:
Ideally we should not store this info in the db but query it from identity provider when displaying. The reason is that the user name could be changed and will be outdated in the db

geofjamg:
Bit maybe if we use the email as user id we don t need to display anything else

src/main/resources/study.cql
```diff
@@ -5,5 +5,7 @@ CREATE TABLE IF NOT EXISTS study.study (
     description text,
     caseFormat text,
     caseUuid uuid,
-    casePrivate boolean
+    casePrivate boolean,
```
geofjamg, line 8: This is the study that is private not only the case

src/main/resources/study.cql
```diff
@@ -5,5 +5,7 @@ CREATE TABLE IF NOT EXISTS study.study (
...
+    ownerEmail text,
+    ownerName text
```
geofjamg, line 10: To remove?

pom.xml
```diff
@@ -164,6 +164,13 @@
             <groupId>com.powsybl</groupId>
             <artifactId>powsybl-network-store-client</artifactId>
             <version>${network-store-client.version}</version>
+            <exclusions>
```
jonenst, line 167: why this ? remove it ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -96,22 +99,39 @@ public StudyService(
...
     }
 
-    Flux<StudyInfos> getStudyList() {
-        Flux<Study> studyList = studyRepository.findAll();
-        return studyList.map(study -> new StudyInfos(study.getName(), study.getDescription(), study.getCaseFormat()));
+    Flux<StudyInfos> getStudyList(String subject) {
+        Flux<StudyBySubject> studyBySubjectFlux = studyBySubjectRepository.findAllBySubject(subject);
+
+        return Flux.concat(studyBySubjectFlux.flatMap(studyBySubject -> {
+            Mono<Study> studyMono = studyRepository.findByName(studyBySubject.getStudyName());
```
jonenst, line 110: I think we should denormalize the data to avoid having to do 1 request per study ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -96,22 +99,39 @@ public StudyService(
...
+            return studyMono.map(study -> new StudyInfos(study.getName(), study.getDescription(), study.getCaseFormat()));
+        }), getPublicStudyList());
+    }
+
+    Flux<StudyInfos> getPublicStudyList() {
+        Flux<StudyByAccessRights> studyByAccessRightsFlux = studyByAccessRightsRepository.findAllByIsPrivate(false);
+        return studyByAccessRightsFlux.flatMap(studyByAccessRights -> {
+            Mono<Study> studyMono = studyRepository.findByName(studyByAccessRights.getStudyName());
```
jonenst, line 118: same remark, denormalize ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -96,22 +99,39 @@ public StudyService(
...
         return Mono.zip(networkInfos, caseFormat)
-                .flatMap(t -> {
-                    Study study = new Study(studyName, t.getT1().getNetworkUuid(), t.getT1().getNetworkId(), description, t.getT2(), caseUuid, false);
-                    return studyRepository.insert(study);
-                });
+            .flatMap(t -> {
+                final StudyByAccessRights studyByAccessRights = new StudyByAccessRights(isPrivate, studyName);
+                final StudyBySubject studyBySubject = new StudyBySubject(subject, studyName);
+                final Study study = new Study(studyName, subject, t.getT1().getNetworkUuid(), t.getT1().getNetworkId(), description, t.getT2(), caseUuid, false, isPrivate);
+                return studyBySubjectRepository.insert(studyBySubject).then(studyByAccessRightsRepository.insert(studyByAccessRights).then(studyRepository.insert(study)));
```
jonenst, line 133: do them in parallel instead of sequentially

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -125,16 +145,24 @@ public StudyService(
...
             Mono<NetworkInfos> networkInfos = persistentStore(uuid);
             Mono<String> caseFormat = getCaseFormat(uuid);
             return Mono.zip(networkInfos, caseFormat)
                     .flatMap(t -> {
-                        Study study = new Study(studyName, t.getT1().getNetworkUuid(), t.getT1().getNetworkId(), description, t.getT2(), uuid, true);
-                        return studyRepository.insert(study);
+                        final StudyBySubject studyBySubject = new StudyBySubject(subject, studyName);
+                        final StudyByAccessRights studyByAccessRights = new StudyByAccessRights(isPrivate, studyName);
+                        final Study study = new Study(studyName, subject, t.getT1().getNetworkUuid(), t.getT1().getNetworkId(), description, t.getT2(), uuid, true, isPrivate);
+                        return studyBySubjectRepository.insert(studyBySubject).then(
```
jonenst, line 161: same remark, parallel instead of sequential

src/main/java/org/gridsuite/study/server/StudyService.java
jonenst: also keep only the public studies not from this user (because they are already included in the other request). Or alternatively, make the other request return only the private studies of the subject.

jonenst (changes requested):
Need to store not only the subject, but also the issuer

jonenst (approved):

jonenst (changes requested):
use "Synthetic Sharding Strategy: Known Shard Count"  for the StudyByAccessRight table

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -96,22 +101,34 @@ public StudyService(
...
+
+    Flux<StudyInfos> getPublicStudyList() {
+        Flux<PublicStudy> publicStudyFlux = publicStudyRepository.findAll();
+        return publicStudyFlux.map(publicStudy ->
+            new StudyInfos(publicStudy.getStudyName(), publicStudy.getUserId(), publicStudy.getDescription(), publicStudy.getCaseFormat())
+        );
     }
 
-    Mono<Study> createStudy(String studyName, UUID caseUuid, String description) {
+    @Transactional
```
jonenst, line 123: not needed since createStudy is @Transactional ?

jonenst: Ah I meant insertStudy. But I think you need this.insertStudy to use the proxy ?

jonenst: Nah, foo() and this.foo() are the same... What you did is best, ignore my remark

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -125,28 +142,57 @@ public StudyService(
...
+        return studyRepository.findByNameAndUserId(studyName, userId);
     }
 
-    Mono<Void> deleteStudy(String studyName) {
-        Mono<Study> studyMono = studyRepository.findByName(studyName);
+    Mono<Void> deleteStudy(String studyName, String userId, String headerUserId) {
+        Mono<Study> studyMono = studyRepository.findByNameAndUserId(studyName, userId);
 
         return studyMono.flatMap(study -> {
+            //if the study is public we need to ensure that it's the initial creator that deletes it
```
jonenst, line 192: need to ensure that also for private studies ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -125,28 +142,57 @@ public StudyService(
...
 
-    Mono<Study> getStudy(String studyName) {
-        return studyRepository.findByName(studyName);
+    Mono<Study> getStudy(String studyName, String userId) {
+        return studyRepository.findByNameAndUserId(studyName, userId);
     }
 
-    Mono<Void> deleteStudy(String studyName) {
-        Mono<Study> studyMono = studyRepository.findByName(studyName);
+    Mono<Void> deleteStudy(String studyName, String userId, String headerUserId) {
```
jonenst, line 188: `@Transactional` ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -156,14 +202,19 @@ public StudyService(
                         .uri(caseServerBaseUri + path)
                         .retrieve()
                         .bodyToMono(Void.class)
-                        .then(studyRepository.deleteByName(studyName));
+                        .then(Mono.zip(studyBySubjectRepository.delete(userId, study.isPrivate(), studyName),
+                                publicStudyRepository.delete(studyName, userId),
+                                studyRepository.deleteByNameAndUserId(studyName, userId)).map(Tuple3::getT3));
```
jonenst, line 207: refactor these 3 lines with the 3 lines below, they are the same

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -334,13 +385,44 @@ public StudyService(
     }
 
     @Transactional
-    public Mono<Study> renameStudy(String studyName, String newStudyName) {
-        Mono<Study> studyMono = studyRepository.findByName(studyName);
+    public Mono<Study> renameStudy(String studyName, String userId, String newStudyName) {
```
jonenst, line 388: Call insert and delete instead of manually redoing the same things ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -334,13 +385,44 @@ public StudyService(
...
+            if (!study.isPrivate()) {
+                Mono<PublicStudy> publicStudyMono = publicStudyRepository.findById(studyName, userId);
+                Mono<PublicStudy> insertPublicStudyMono = Mono.zip(publicStudyMono, studyBySubjectMono).flatMap(t -> {
+                    PublicStudy publicStudy = t.getT1();
+                    publicStudy.setStudyName(newStudyName);
+
+                    StudyBySubject studyBySubject = t.getT2();
+                    studyBySubject.setStudyName(newStudyName);
+
+                    return studyBySubjectRepository.insert(studyBySubject).then(publicStudyRepository.insert(publicStudy));
```
jonenst, line 404: should be in parallel, not sequential

pom.xml
```diff
@@ -155,6 +155,11 @@
 
     <dependencies>
         <!-- Compilation dependencies -->
+        <dependency>
```
jonenst, line 158: why cassandra-driver-extras ?

src/main/resources/study.cql
```diff
@@ -1,9 +1,38 @@
...
     networkUuid uuid,
     networkId text,
     description text,
     caseFormat text,
     caseUuid uuid,
-    casePrivate boolean
+    casePrivate boolean,
+    userId text,
+    isPrivate boolean,
+    PRIMARY KEY (studyName, userId)
```
jonenst, line 11: At least, the primary key should be userId, studyName.

Maybe for now the simplest solution is to do

CREATE TABLE IF NOT EXISTS privateStudy
     PRIMARY KEY (userId, studyName) 

CREATE TABLE IF NOT EXISTS publicStudy
     PRIMARY KEY (userId, studyName) 

I believe these two tables are enough for all our queries.

jonenst:
Also remove all references to Subject and use UserId or User instead please

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -156,14 +203,21 @@ public StudyService(
...
+                return deleteRequests(userId, studyName);
             }
         });
     }
 
+    Mono<Void> deleteRequests(String userId, String studyName) {
+        return Mono.zip(privateStudyBySubjectRepository.delete(userId, studyName),
+                publicStudyRepository.delete(studyName, userId),
+                studyRepository.deleteByNameAndUserId(studyName, userId)).map(Tuple3::getT3);
+    }
```
jonenst, line 217: use .then() instead of .map(getT3) ?

src/main/java/org/gridsuite/study/server/RestResponseEntityExceptionHandler.java
```diff
@@ -19,6 +19,8 @@
             return ResponseEntity.status(HttpStatus.FAILED_DEPENDENCY).body(CASE_DOESNT_EXISTS);
         } else if (errorMessage.equals(STUDY_ALREADY_EXISTS)) {
             return ResponseEntity.status(HttpStatus.CONFLICT).body(STUDY_ALREADY_EXISTS);
+        } else if (errorMessage.equals(NOT_ALLOWED)) {
+            return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body(NOT_ALLOWED);
```
jonenst, line 23: should be  403 Forbidden  instead of 401 Unauthorized

src/main/java/org/gridsuite/study/server/repository/PublicStudyRepository.java
```diff
@@ -0,0 +1,22 @@
+package org.gridsuite.study.server.repository;
```
jonenst, line 1: missing copyright

src/main/java/org/gridsuite/study/server/repository/PublicStudy.java
```diff
@@ -0,0 +1,52 @@
+package org.gridsuite.study.server.repository;
```
jonenst, line 1: I think we can share the StudyEntity between the 3 repositories ?

AbdelHedhili: wa can't the name of the class references the name of the table

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -334,13 +387,19 @@ public StudyService(
...
     @Transactional
-    public Mono<Study> renameStudy(String studyName, String newStudyName) {
-        Mono<Study> studyMono = studyRepository.findByName(studyName);
+    public Mono<Study> renameStudy(String studyName, String userId, String headerUserId, String newStudyName) {
+        if (!userId.equals(headerUserId)) {
+            return Mono.error(new StudyException(NOT_ALLOWED));
+        }
+        Mono<Study> studyMono = studyRepository.findByUserIdAndName(userId, studyName);
         return studyMono.switchIfEmpty(Mono.error(new StudyException(STUDY_DOESNT_EXISTS))).flatMap(study -> {
             study.setName(newStudyName);
```
jonenst, line 396: remove this line ?

jonenst: not sure..

jonenst: yeah no keep it

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -125,28 +141,58 @@ public StudyService(
...
+        if (!isPrivate) {
+            return Mono.zip(publicStudyRepository.insert(publicStudy), studyRepository.insert(study))
+                    .map(Tuple2::getT2);
+        } else {
+            return Mono.zip(privateStudyRepository.insert(privateStudy), studyRepository.insert(study))
+                    .map(Tuple2::getT2);
+        }
+    }
+
+    Mono<Study> getUserStudy(String studyName, String userId, String headerUserId) {
```
jonenst, line 172: rename to getCurrentUserStudy ?

jonenst (approved):

PR #56, opened 2020-11-17
convert studyname for update message

JacquesBorsenberger:
Signed-off-by: jacqbors <jacques.borsenberger@rte-france.com>


jonenst (changes requested):
We shouldn't send urlencoded stuff to rabbitmq which has nothing to do with browsers and urls.
Instead, if we need to url encode things, we need to do it in the client, where we put this name in the browser url.

PR #57, opened 2020-11-18
Store security analysis status

FranckLecuyer:


src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -640,7 +645,9 @@ private void emitStudyChanged(String studyName, String updateType) {
     }
 
     private Mono<Void> assertSecurityAnalysisNotRunning(String studyName, String userId) {
-        return Mono.empty(); // FIXME the security analysis status is not yet implemented
+        Mono<String> statusMono = getSecurityAnalysisStatus(studyName, userId);
+        return statusMono.switchIfEmpty(Mono.empty())
```
jonenst, line 649: switchIfEmpty(Mono.empty()) doesn't do anthing ??

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -640,7 +645,9 @@ private void emitStudyChanged(String studyName, String updateType) {
...
+                .flatMap(s -> s.equals(SecurityAnalysisStatus.RUNNING.name()) ? Mono.error(new StudyException(SECURITY_ANALYSIS_RUNNING)) : Mono.empty());
```
jonenst, line 650: When can statusMono be empty ? (warning: flatmap doesn't execute on empty monos)

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -749,6 +759,12 @@ public static ComponentResultEntity toEntity(LoadFlowResult.ComponentResult comp
                     .retrieve()
                     .bodyToMono(UUID.class);
         });
+
+        return resultUuid.flatMap(result ->
+                  studyRepository.updateSecurityAnalysisResultUuid(studyName, userId, result)
+                .doOnSuccess(ignored -> emitStudyChanged(studyName, StudyService.UPDATE_TYPE_SECURITY_ANALYSIS_STATUS))
```
jonenst, line 765: use the same convention as the rest of the file:
  .doOnSuccess(e -> ...)

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -749,6 +759,12 @@ public static ComponentResultEntity toEntity(LoadFlowResult.ComponentResult comp
...
+                         .then(Mono.just(result))
```
jonenst, line 766: .thenReturn(result)

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -729,7 +738,7 @@ public static ComponentResultEntity toEntity(LoadFlowResult.ComponentResult comp
 
         Mono<UUID> networkUuid = getNetworkUuid(studyName, userId);
 
-        return networkUuid.flatMap(uuid -> {
+        Mono<UUID> resultUuid = networkUuid.flatMap(uuid -> {
```
jonenst, line 741: Keep the previous style, don't create a new variable and just chain your flatmap to this returned expression

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -833,6 +849,46 @@ public static ComponentResultEntity toEntity(LoadFlowResult.ComponentResult comp
                 .bodyToMono(String.class);
     }
 
+    public Mono<String> getSecurityAnalysisStatus(String studyName, String userId) {
+        Objects.requireNonNull(studyName);
+        Objects.requireNonNull(userId);
+
+        return studyRepository.findStudy(userId, studyName).flatMap(entity -> {
+            UUID resultUuid = entity.getSecurityAnalysisResultUuid();
+            return Mono.justOrEmpty(resultUuid).flatMap(uuid -> {
```
jonenst, line 858: when can resultUuid be empty ? warning: flatmap doesn't execute in this case

jonenst: Shouldn't it error when the uuid is null, instead of returning an empty mono ?

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -833,6 +849,46 @@ public static ComponentResultEntity toEntity(LoadFlowResult.ComponentResult comp
...
+            });
+        });
+    }
+
+    public Mono<Void> setSecurityAnalysisStatus(String studyName, String userId, SecurityAnalysisStatus status) {
+        Objects.requireNonNull(studyName);
+        Objects.requireNonNull(userId);
+
+        return studyRepository.findStudy(userId, studyName).flatMap(entity -> {
+            UUID resultUuid = entity.getSecurityAnalysisResultUuid();
```
jonenst, line 877: same:



when can resultUuid be empty ? warning: flatmap doesn't execute in this case

Shouldn't it error when the uuid is null, instead of returning an empty mono ?

src/main/java/org/gridsuite/study/server/StudyService.java
FranckLecuyer: Corrected

src/main/java/org/gridsuite/study/server/StudyService.java
FranckLecuyer: Corrected

src/main/java/org/gridsuite/study/server/StudyService.java
FranckLecuyer: Corrected

src/main/java/org/gridsuite/study/server/StudyService.java
FranckLecuyer: Corrected

src/main/java/org/gridsuite/study/server/StudyService.java
FranckLecuyer: Corrected

src/main/java/org/gridsuite/study/server/StudyService.java
geofjamg: Agree, if no security analysis result in the DB, it means the API is not used correctly and we should throw an error.

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -833,6 +848,46 @@ public static ComponentResultEntity toEntity(LoadFlowResult.ComponentResult comp
...
+                        .get()
+                        .uri(securityAnalysisServerBaseUri + path)
+                        .retrieve()
+                        .onStatus(httpStatus -> httpStatus == HttpStatus.NOT_FOUND, clientResponse -> Mono.error(new StudyException(SECURITY_ANALYSIS_NOT_FOUND)))
+                        .bodyToMono(String.class);
+            });
+        });
+    }
+
+    public Mono<Void> setSecurityAnalysisStatus(String studyName, String userId, SecurityAnalysisStatus status) {
```
geofjamg, line 871: To remove, we don't want to set the result status from API, it is supposed to be a computation state.

geofjamg (changes requested):

src/main/java/org/gridsuite/study/server/StudyService.java
FranckLecuyer: Here, it's the same thing as getting the security analysis result : it can be called on study creation, when no security analysis has been launched ...

geofjamg: Ok

src/main/java/org/gridsuite/study/server/StudyService.java
FranckLecuyer: Changed to 'invalidateSecurityAnalysisStatus'

PR #85, opened 2021-04-28
 Fix notifications for private studies to send them only to owner

SlimaneAmar:
Fix notifications for private studies to send them only to owner and not to all users

### DIFF
--- a/src/main/java/org/gridsuite/study/server/StudyService.java
+++ b/src/main/java/org/gridsuite/study/server/StudyService.java
@@ -242,7 +242,7 @@ public class StudyService {
     public Mono<BasicStudyInfos> createStudy(String studyName, UUID caseUuid, String description, String userId, Boolean isPrivate) {
         return insertStudyCreationRequest(studyName, userId, isPrivate)
                 .map(StudyService::toBasicStudyInfos)
-                .doOnSuccess(s -> Mono.zip(persistentStore(caseUuid, studyName, userId, isPrivate), getCaseFormat(caseUuid))
+                .doOnSuccess(s -> Mono.zip(persistentStore(caseUuid, s.getStudyUuid(), studyName, userId, isPrivate), getCaseFormat(caseUuid))
                         .flatMap(t -> {
                             LoadFlowParameters loadFlowParameters = LoadFlowParameters.load();
                             return insertStudy(s.getStudyUuid(), studyName, userId, isPrivate, t.getT1().getNetworkUuid(), t.getT1().getNetworkId(),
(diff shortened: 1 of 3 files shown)


JacquesBorsenberger (approved):

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -400,25 +403,23 @@ public StudyEntity doGetStudyWithPreFetchedLoadFlowResultAndUpdateIsPrivate(UUID
...
                 JsonNode node = new ObjectMapper().readTree(body).path("message");
                 if (!node.isMissingNode()) {
-                    message = node.asText();
-                    emitStudyError(studyName, UPDATE_TYPE_STUDIES, message);
+                    emitStudyCreationError(studyName, userId, isPrivate, node.asText());
                 }
             } catch (JsonProcessingException e) {
                 if (!body.isEmpty()) {
-                    emitStudyError(studyName, UPDATE_TYPE_STUDIES, body);
+                    emitStudyCreationError(studyName, userId, isPrivate, body);
```
jonenst, line 415: maybe in another PR and in another repo, but I think we should not send error notifications to other users for public studies.

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -769,10 +780,12 @@ private void emitStudyChanged(UUID studyUuid, String updateType) {
         );
     }
 
-    private void emitStudyError(String studyName, String updateType, String errorMessage) {
+    private void emitStudyCreationError(String studyName, String userId, boolean isPrivate, String errorMessage) {
         studyUpdatePublisher.onNext(MessageBuilder.withPayload("")
                 .setHeader(HEADER_STUDY_NAME, studyName)
-                .setHeader(HEADER_UPDATE_TYPE, updateType)
```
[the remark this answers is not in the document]
SlimaneAmar: Done

src/main/java/org/gridsuite/study/server/StudyService.java
SlimaneAmar: Done

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -400,25 +403,23 @@ public StudyEntity doGetStudyWithPreFetchedLoadFlowResultAndUpdateIsPrivate(UUID
...
-    private Mono<? extends Throwable> handleStudyCreationError(String studyName, ClientResponse clientResponse) {
+    private Mono<? extends Throwable> handleStudyCreationError(UUID studyUuid, String studyName, String userId, ClientResponse clientResponse) {
         return clientResponse.bodyToMono(String.class).flatMap(body -> {
             try {
-                String message;
                 JsonNode node = new ObjectMapper().readTree(body).path("message");
                 if (!node.isMissingNode()) {
-                    message = node.asText();
-                    emitStudyError(studyName, UPDATE_TYPE_STUDIES, message);
+                    emitStudyCreationError(studyUuid, studyName, userId, true, node.asText());
```
jonenst, line 411: if this is force to true here, then it should not be called "publicstudy", it should be called "broadcastnotification" or something like that.
But instead, I think we should send the real value of isPrivate, and add filter in the notification-server

jonenst (approved):

src/main/java/org/gridsuite/study/server/StudyService.java
SlimaneAmar: Done

PR #95, opened 2021-06-11
Add single equipment request

JacquesBorsenberger:


src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -540,4 +546,17 @@ public StudyController(StudyService studyService) {
...
+    @GetMapping(value = "/studies/{studyUuid}/network-map/{equipmentType}/{equipmentId}")
+    @ApiOperation(value = "Get Network equipment description", produces = "application/json")
+    @ApiResponses(value = {@ApiResponse(code = 200, message = "The equipment data")})
+    public ResponseEntity<Mono<String>> getEquipmentData(
+        @PathVariable("studyUuid") UUID studyUuid,
+        @PathVariable("equipmentType") EquipmentType equipmentType,
+        @PathVariable("equipmentId") String equipmentId) {
+
+        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studyService.getNetworkUuid(studyUuid)
+            .flatMap(uuid -> studyService.getEquipmentData(uuid, equipmentType, equipmentId)));
```
niconoir, line 559: Here, equipmentType can be null, since the PathVariable "equipmentType" can be different from the accepted types of the enum (method fromString of the EquipmentType returns null if type is not one of the enum, and so EquipmentTypeEnumConverter setAsText method sets a null value). The case of a null value for equipmentType is not correctly addressed in the getEquipmentData method of the StudyService.

src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -556,6 +556,15 @@ public StudyEntity doGetStudyWithPreFetchedLoadFlowResultAndUpdateIsPrivate(UUID
                 .bodyToMono(String.class);
     }
 
+    public Mono<String> getEquipmentData(UUID networkUuid, EquipmentType equipmentType, String equipmentId) {
+        UriComponentsBuilder builder = UriComponentsBuilder.fromPath(DELIMITER + NETWORK_MAP_API_VERSION + "/networks/{networkUuid}/{equipmentPath}/{equipmentId}");
+        String path = builder.buildAndExpand(networkUuid, equipmentType.getServerPath(), equipmentId).toUriString();
```
niconoir, line 561: equipmentType can be null here (risk of NullPointerException)

src/test/java/org/gridsuite/study/server/StudyTest.java
```diff
@@ -349,6 +350,8 @@ public MockResponse dispatch(RecordedRequest request) throws InterruptedExceptio
                     case "/v1/substations/38400000-8cf0-11bd-b23e-10b96e4ef00d":
                     case "/v1/2-windings-transformers/38400000-8cf0-11bd-b23e-10b96e4ef00d":
                     case "/v1/3-windings-transformers/38400000-8cf0-11bd-b23e-10b96e4ef00d":
+                    case "/v1/networks/38400000-8cf0-11bd-b23e-10b96e4ef00d/2-windings-transformers/" + EQUIPMENT_UUID:
+                    case "/v1/networks/38400000-8cf0-11bd-b23e-10b96e4ef00d/3-windings-transformers/" + EQUIPMENT_UUID:
```
niconoir, line 354: Why not use NETWORK_UUID_STRING here ?

src/test/java/org/gridsuite/study/server/StudyTest.java
```diff
@@ -1157,6 +1160,24 @@ public void testDiagramsAndGraphics() {
...
+
+        //get the 3 windings transformers map data of a network
+        webTestClient.get()
+            .uri("/v1/studies/{studyUuid}/network-map/3-windings-transformers/{tranformerUUID}", studyNameUserIdUuid, EQUIPMENT_UUID)
+            .exchange()
+            .expectStatus().isOk()
+            .expectHeader().contentType(MediaType.APPLICATION_JSON);
+
+        assertTrue(getRequestsDone(1).contains(String.format("/v1/networks/%s/3-windings-transformers/%s", NETWORK_UUID_STRING, EQUIPMENT_UUID)));
+
```
niconoir, line 1180: Maybe add a test with a wrong equipment type ?

niconoir (changes requested):

src/main/java/org/gridsuite/study/server/StudyController.java
JacquesBorsenberger: nope : you get the following error with a bad type
org.springframework.web.server.ServerErrorException: 500 INTERNAL_SERVER_ERROR "equipmentType"

niconoir: Maybe throw a more explicit exception then ?

PR #102, opened 2021-07-09
[WIP] manage studies with directory

AbdelHedhili:
Signed-off-by: AbdelHedhili <abdelsalem.hedhili@rte-france.com>


AbdelHedhili:
We no longer want the study-server as an entry point so this PR is not needed anymore

PR #131, opened 2021-11-08
Store voltage level name in elasticsearch

niconoir:


src/main/java/org/gridsuite/study/server/dto/VoltageLevelInfos.java
```diff
@@ -20,6 +21,7 @@
 @NoArgsConstructor
 @Getter
 @ToString
+@EqualsAndHashCode
```
sBouzols, line 24: Alternatively, you can specify exactly which fields or methods you wish to be used by marking them with @EqualsAndHashCode.Include and using @EqualsAndHashCode(onlyExplicitlyIncluded = true). 

Then maybe only on the `'id'` attribute ?

sBouzols (approved):
rought Code review OK
a question about `EqualsAndHashCode`
No tests execution

SlimaneAmar (approved):

PR #153, opened 2022-01-14
keep data on node update

JacquesBorsenberger:
Signed-off-by: jacques borsenberger <jacques.borsenberger@rte-france.com>


niconoir (approved):
Not tested

PR #176, opened 2022-03-08
don't delete and reinsert loadflowresults when invalidating the loadflow status

jonenst:
the previous method was ignoring null fields, so removing the null LoadFlowResult
and calling updateLoadFlowStatus should be equivalent. Avoiding a round trip through
dtos allows to avoid unecessary deletes and inserts.

Signed-off-by: HARPER Jon <jon.harper87@gmail.com>


laurgarn (approved):

src/main/java/org/gridsuite/study/server/networkmodificationtree/ModelNodeInfoRepositoryProxy.java
```diff
@@ -72,6 +75,11 @@ public void updateLoadFlowResultAndStatus(AbstractNode node, LoadFlowResult load
         updateNode(modificationNode);
     }
 
+    @Override
+    public void updateLoadFlowStatus(UUID nodeUuid, LoadFlowStatus loadFlowStatus) {
```
geofjamg, line 79: So here you override impl of AbstractNodeRepositoryProxy ?
Why not removing the useless code in AbstractNodeRepositoryProxy ?
Would the best fix be to fix void updateLoadFlowStatus(AbstractNode node, LoadFlowStatus loadFlowStatus) instead to not replace the node when just updating the status?

geofjamg (changes requested):

src/main/java/org/gridsuite/study/server/networkmodificationtree/ModelNodeInfoRepositoryProxy.java
geofjamg:
```java

    @Override
    public void updateLoadFlowStatus(AbstractNode node, LoadFlowStatus loadFlowStatus) {
        ModelNode modelNode = (ModelNode) node;
        modelNode.setLoadFlowStatus(loadFlowStatus);
        updateNode(modelNode);
    }
```

geofjamg: Why do we update the node here?

PR #189, opened 2022-04-21
get network area diagram

etiennehomer:
Signed-off-by: Etienne Homer <etiennehomer@gmail.com>


src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -647,6 +647,17 @@ public ResponseEntity<Mono<String>> getSubstationDiagramAndMetadata(
             new DiagramParameters(useName, centerLabel, diagonalLabel, topologicalColoring, componentLibrary), substationLayout, nodeUuid));
     }
 
+    @GetMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/network/network-area-diagram")
```
geofjamg, line 650: /studies/{studyUuid}/nodes/{nodeUuid}/network-area-diagram instead? network is useless here

geofjamg (changes requested):

src/main/java/org/gridsuite/study/server/StudyController.java
etiennehomer: Yes, done

geofjamg (approved):

PR #193, opened 2022-04-26
Loadflow provider default value can now be defined in application.properties

archendel:


src/main/resources/application-local.yml
```diff
@@ -32,3 +32,7 @@ backing-services:
 
 network-store-server:
   base-uri: http://localhost:8080/
+
+# default values for loadflow providers "Hades2" or "OpenLoadFlow"
+loadflow:
+  default-provider: OpenLoadFlow
```
jonenst, line 38: should not be set here ?

archendel: property moved to default application.yml

jonenst (approved):

PR #199, opened 2022-05-11
No build invalidation of children nodes when building a node

FranckLecuyer:


src/test/java/org/gridsuite/study/server/StudyTest.java
```diff
@@ -2882,20 +2882,19 @@ public void testBuild() {
         // build modificationNode2 and stop build
         testBuildWithNodeUuid(studyNameUserIdUuid, modificationNode2.getId());
 
-        assertEquals(BuildStatus.NOT_BUILT, networkModificationTreeService.getBuildStatus(modificationNode2.getId()));
```
niconoir, line 2885: You can keep this assert in order to assert previous modificationNode2 building was correctly stopped. This behaviour is not modified by this PR.

niconoir: Node status NOT_BUILT is already verified at the end of testBuildWithNodeUuid method, so no need to verify here. assert can be removed, my bad.

niconoir (approved):

PR #206, opened 2022-05-19
Fix pb when getting modifications on node

FranckLecuyer:


SlimaneAmar:
I think it's dangerous to have this method with `generateId` = true by default
`    public Mono<NodeModificationInfos> getNodeModificationInfos(UUID nodeUuid) {
        return Mono.fromCallable(() -> self.doGetNodeModificationInfos(nodeUuid, true))
            .switchIfEmpty(Mono.error(new StudyException(ELEMENT_NOT_FOUND)));
    }`

It's better to always call getNodeModificationInfos with `generateId` or set false by default
Another US ?

SlimaneAmar (approved):

PR #210, opened 2022-06-01
Fix : need to save "running" LF status in database before running LF

SlimaneAmar:
Signed-off-by: Slimane AMAR <slimane.amar@rte-france.com>


archendel (approved):
Tested before and after fix, loadflow status is now correctly updated and displayed on front end

PR #217, opened 2022-06-13
Fixed an exception happening randomly because of a multipart file that was deleted at the end of a request but was read asynchronously

archendel:


src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -303,12 +307,23 @@ private void createStudyAsync(UUID caseUuid, String userId, BasicStudyInfos basi
...
-        studyServerExecutionService.runAsync(() -> createStudyAsync(caseFile, userId, basicStudyInfos));
+        // Using temp file to store caseFile here because multipartfile are deleted once the request using it is over
+        // Since the next action is asynchronous, the multipartfile could be deleted before being read and cause exceptions
+        File tempFile;
+        try {
+            tempFile = File.createTempFile("tmp_", caseFile.getOriginalFilename());
+            caseFile.transferTo(tempFile);
+        } catch (IOException e) {
+            LOGGER.error(e.toString(), e);
+            throw new StudyException(STUDY_CREATION_FAILED);
```
SlimaneAmar, line 318: ` emitStudyCreationError(studyUuid, userId, e.getMessage()); `?

archendel: Is it mandatory since the initial API call would already return an error ?

SlimaneAmar (changes requested):

SlimaneAmar (approved):

PR #221, opened 2022-06-20
improve error message on modification blocked

etiennehomer:
Signed-off-by: Etienne Homer <etiennehomer@gmail.com>

### DIFF
--- a/src/main/java/org/gridsuite/study/server/StudyService.java
+++ b/src/main/java/org/gridsuite/study/server/StudyService.java
@@ -1216,7 +1216,7 @@ public class StudyService {
     public void assertNoNodeIsBuilding(UUID studyUuid) {
         networkModificationTreeService.getAllNodes(studyUuid).stream().forEach(node -> {
             if (networkModificationTreeService.getBuildStatus(node.getIdNode()) == BuildStatus.BUILDING) {
-                throw new StudyException(NOT_ALLOWED);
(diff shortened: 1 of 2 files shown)


src/main/java/org/gridsuite/study/server/StudyService.java
```diff
@@ -1216,7 +1216,7 @@ public void assertNoBuildNoComputation(UUID studyUuid, UUID nodeUuid) {
     public void assertNoNodeIsBuilding(UUID studyUuid) {
         networkModificationTreeService.getAllNodes(studyUuid).stream().forEach(node -> {
             if (networkModificationTreeService.getBuildStatus(node.getIdNode()) == BuildStatus.BUILDING) {
-                throw new StudyException(NOT_ALLOWED);
+                throw new StudyException(NOT_ALLOWED, "The modification node \"" + networkModificationTreeService.getSimpleNode(node.getIdNode()).getName() + "\" is building.");
```
niconoir: Is it necessary to get the name of the currently building node, which implies another database request just to get the node name ?

niconoir (approved):

src/main/java/org/gridsuite/study/server/StudyService.java
etiennehomer: We can make it simplier yes. This error should never be displayed as modifications during building are block in the front.
I changed it.

PR #258, opened 2022-08-01
Synchronize notifications with transaction completion

niconoir:


src/main/java/org/gridsuite/study/server/NotificationService.java
```diff
@@ -0,0 +1,187 @@
...
+
+    static final String HEADER_NODES = "nodes";
+    static final String HEADER_PARENT_NODE = "parentNode";
+    static final String HEADER_NEW_NODE = "newNode";
+    static final String HEADER_REMOVE_CHILDREN = "removeChildren";
+    static final String NODE_UPDATED = "nodeUpdated";
+    static final String NODE_DELETED = "nodeDeleted";
+    static final String NODE_CREATED = "nodeCreated";
+    static final String HEADER_INSERT_MODE = "insertMode";
+
```
SlimaneAmar, line 50: Move here MODIFICATIONS_*_IN_PROGRESS notifications

src/main/java/org/gridsuite/study/server/NotificationService.java
```diff
@@ -0,0 +1,187 @@
...
+    static final String HEADER_NODE = "node";
+    static final String HEADER_UPDATE_TYPE = "updateType";
+    static final String UPDATE_TYPE_STUDIES = "studies";
+    static final String UPDATE_TYPE_STUDY_DELETE = "deleteStudy";
+    static final String HEADER_ERROR = "error";
+    static final String HEADER_UPDATE_TYPE_SUBSTATIONS_IDS = "substationsIds";
+    static final String HEADER_UPDATE_TYPE_DELETED_EQUIPMENT_ID = "deletedEquipmentId";
+    static final String HEADER_UPDATE_TYPE_DELETED_EQUIPMENT_TYPE = "deletedEquipmentType";
+
+    static final String HEADER_NODES = "nodes";
```
SlimaneAmar, line 42: Use an enum for notification types ?

SlimaneAmar (changes requested):

src/main/java/org/gridsuite/study/server/NotificationService.java
niconoir: All notifications types moved in NotificationService

src/main/java/org/gridsuite/study/server/NotificationService.java
niconoir: Using enum for notification types would add complexity to the code. Since we only need String to pass as parameters to the MessageBuilder (no value checking needed), static final String seem sufficient.

SlimaneAmar (approved):

PR #295, opened 2022-11-07
fixing clean of data base

anas-drissi:


src/test/java/org/gridsuite/study/server/service/ReportServiceTest.java
```diff
@@ -143,6 +148,7 @@ public MockResponse dispatch(RecordedRequest request) {
     public void tearDown() {
         TestUtils.assertQueuesEmptyThenClear(List.of(STUDY_UPDATE_DESTINATION), output);
         try {
+            cleanDB();
```
niconoir, line 151: Maybe call cleanDB at the beginning of the tearDown method to ensure database is cleaned even if TestUtils.assertQueuesEmptyThenClear fails (to ensure other tests will work correctly) or add a try ... finally block on the whole method and clean the db in the finally block.

niconoir (changes requested):

niconoir (approved):

PR #296, opened 2022-11-10
Use the study loadflow parameters for the security analysis

FranckLecuyer:


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -969,15 +970,29 @@ public UUID runSecurityAnalysis(UUID studyUuid, List<String> contingencyListName
...
+                StandardCharsets.UTF_8);
         } catch (JsonProcessingException e) {
             throw new UncheckedIOException(e);
         }
 
         Optional<UUID> prevResultUuidOpt = networkModificationTreeService.getSecurityAnalysisResultUuid(nodeUuid);
         prevResultUuidOpt.ifPresent(securityAnalysisService::deleteSaResult);
 
-        UUID result = securityAnalysisService.runSecurityAnalysis(networkUuid, reportUuid, nodeUuid, variantId, provider, contingencyListNames, parameters, receiver);
+        SecurityAnalysisParameters securityAnalysisParameters = new SecurityAnalysisParameters();
```
geofjamg, line 981: As we do not manage security analysis parameters persistence in study (not yet), maybe it should be better if no loadflowparameters stored to not send security analysis parameters and let security analysis server decide which are the default one

geofjamg (approved):

PR #299, opened 2022-11-17
Moving a node around itself now throws an exception

archendel:


AbdelHedhili (approved):

AbdelHedhili (changes requested):
while you are at it add a test for this case

AbdelHedhili (approved):

bendaamerahm (approved):

PR #305, opened 2022-11-24
add displayMode as  new param to sld (voltage-level) api

bendaamerahm:
add displayMode as new param to sld (voltage-level) api


geofjamg (changes requested):
Instead of useFeederPositions we should have an enum so that it could be extended to other kind of display easily: 
something like displayMode:
 - StateVariable
 - FeederPosition

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -59,8 +59,7 @@ public StudyController(StudyService studyService,
             NetworkConversionService networkConversionService,
             SecurityAnalysisService securityAnalysisService,
             SensitivityAnalysisService sensitivityAnalysisService,
-            ShortCircuitService shortCircuitService,
-            CaseService caseService) {
+            ShortCircuitService shortCircuitService, CaseService caseService) {
```
etiennehomer, line 62: undo

src/main/java/org/gridsuite/study/server/StudyConstants.java
```diff
@@ -39,4 +39,9 @@ private StudyConstants() {
     public static final String QUERY_PARAM_ERROR_ON_GROUP_NOT_FOUND = "errorOnGroupNotFound";
     public static final String QUERY_PARAM_ERROR_ON_REPORT_NOT_FOUND = "errorOnReportNotFound";
     public static final String QUERY_PARAM_REPORT_DEFAULT_NAME = "defaultName";
+
+    public enum DisplayMode {
```
etiennehomer, line 43: Maybe use a more specific name like SldLabellingMode ? SldDisplayMode?

etiennehomer (changes requested):

src/main/java/org/gridsuite/study/server/StudyConstants.java
bendaamerahm: i agree

geofjamg (approved):

etiennehomer (approved):

PR #314, opened 2022-12-14
add two new types of modifications

bendaamerahm:


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1501,7 +1501,8 @@ public UUID runShortCircuit(UUID studyUuid, UUID nodeUuid) {
 
     private ModificationType getModificationType(String modificationAttributes) {
         try {
-            return objectMapper.readValue(modificationAttributes, ModificationInfos.class).getType();
+            var x = objectMapper.readValue(modificationAttributes, ModificationInfos.class).getType();
+            return x;
```
dbraquart, line 1505: why ?

dbraquart (commented):
sounds OK
could not create "delete attaching line" modification (maybe because of the front - to be confirmed later)

dbraquart (approved):

flomillot (approved):

PR #322, opened 2022-12-23
Fix sld opening pb

FranckLecuyer:


src/main/java/org/gridsuite/study/server/service/NetworkService.java
```diff
@@ -96,4 +98,10 @@ UUID getNetworkUuid(Network network) {
     List<VariantInfos> getNetworkVariants(UUID networkUuid) {
         return networkStoreService.getVariantsInfos(networkUuid).stream().sorted(Comparator.comparing(VariantInfos::getNum)).collect(Collectors.toList());
     }
+
+    boolean existeVariant(UUID networkUuid, String variantId) {
```
EstherDarkish, line 102: Shouldn't `existeVariant` be instead named `existVariant` ?

EstherDarkish (approved):

src/main/java/org/gridsuite/study/server/service/NetworkService.java
FranckLecuyer: Renamed

PR #337, opened 2023-01-25
WIP add branch network-map end-point

dbraquart:


src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -1297,4 +1297,16 @@ public void setAsText(final String text) throws IllegalArgumentException {
             setValue(ModificationType.getTypeFromUri(text));
         }
     }
+
+    @GetMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/network-map/branches/{branchId}")
```
etiennehomer, line 1301: Do we need a new endpoint in study-server and network-map-server or we can do this functionnality with get2WTMapData and getLineMapData ?

dbraquart: it's better a single endpoint because it must be quick.
but waybe I can get everything I want from get2WTMapData, with a small modification.
I'll check this

dbraquart: yes I will try with existing end-points (slightly modified)

PR #346, opened 2023-02-15
add new end-point to retrieve overloaded lines (current limit violations)

dbraquart:


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1106,6 +1110,27 @@ public Integer getContingencyCount(UUID studyUuid, List<String> contingencyListN
...
+                .limitName(violation.getLimitName())
+                .value(violation.getValue())
+                .side(violation.getSide() != null ? violation.getSide().name() : "").build();
+    }
+
+    public List<LimitViolationInfos> getOverloadedLines(UUID studyUuid, UUID nodeUuid, float limitReduction) {
+        Objects.requireNonNull(studyUuid);
+        Objects.requireNonNull(nodeUuid);
+        UUID networkUuid = networkStoreService.getNetworkUuid(studyUuid);
+        Network network = networkStoreService.getNetwork(networkUuid, PreloadingStrategy.ALL_COLLECTIONS_NEEDED_FOR_BUS_VIEW, networkModificationTreeService.getVariantId(nodeUuid));
```
FranckLecuyer, line 1127: Are you sure that ALL_COLLECTIONS_NEEDED_FOR_BUS_VIEW preloading strategy is necessary ?
if Security.checkLimits only check for lines and transformers limits, COLLECTIONS strategy should be enough, no ?

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/study/server/service/StudyService.java
dbraquart: yes COLLECTIONS is what I need

FranckLecuyer (approved):

Meklo (approved):

PR #354, opened 2023-03-06
Dynamic Simulation - Config parameters startTime, stopTime, mapping, solver and provider

thangqp:


src/main/java/org/gridsuite/study/server/repository/DynamicSimulationParametersEntity.java
```diff
@@ -0,0 +1,43 @@
...
+@Setter
+@Entity
+@Table(name = "dynamicSimulationParameters")
+public class DynamicSimulationParametersEntity {
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    @Column(name = "id")
+    private UUID id;
+
+    @Column(name = "startTime", columnDefinition = "numeric default 0")
```
jonenst, line 29: this doesn't look like our other entities

src/main/java/org/gridsuite/study/server/dto/dynamicsimulation/solver/SolverInfos.java
```diff
@@ -0,0 +1,60 @@
...
+        visible = true
+)@JsonSubTypes({
+    @JsonSubTypes.Type(value = IdaSolverInfos.class, name = "IDA"),
+    @JsonSubTypes.Type(value = SimSolverInfos.class, name = "SIM")})
+public interface SolverInfos {
+    String getId();
+
+    SolverTypeInfos getType();
+
+    static List<SolverInfos> parseJson(String json) {
```
jonenst, line 37: this is not like our other json serialization/deserialization, we usually register the modules in springboot (using @configuration in main class)

jonenst (approved):

src/main/java/org/gridsuite/study/server/repository/DynamicSimulationParametersEntity.java
```diff
@@ -0,0 +1,43 @@
...
+    private int startTime;
+
+    @Column(name = "stopTime", columnDefinition = "numeric default 0")
+    private int stopTime;
+
+    @Column(name = "solverId", columnDefinition = "varchar(2) default null")
```
jonenst, line 35: take more margin, varchar(2) but what if a new simulator comes and is longer? This is cheap we don't need to optimize so much

thangqp: Done, set varchar(255)

src/main/java/org/gridsuite/study/server/repository/DynamicSimulationParametersEntity.java
thangqp: Done! now startTime, stopTime in double (demanded by Gautier).. removed columnDefinition = "numeric default 0"

src/main/java/org/gridsuite/study/server/dto/dynamicsimulation/solver/SolverInfos.java
thangqp: To discuss later..

At this moment, this Dto conformes to the standard parsed by default ObjectMapper of jackson.. In my opinion, it is not necessary to write a  prop serialization/deserialization.

PR #359, opened 2023-03-20
reindex a duplicated Study

dbraquart:


laurgarn (approved):
Test : OK
Code : OK

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -275,6 +275,7 @@ private void duplicateStudyAsync(BasicStudyInfos basicStudyInfos, StudyEntity so
             LoadFlowParameters newLoadFlowParameters = sourceLoadFlowParameters != null ? sourceLoadFlowParameters.copy() : new LoadFlowParameters();
             ShortCircuitParameters shortCircuitParameters = copiedShortCircuitParameters != null ? copiedShortCircuitParameters : ShortCircuitService.getDefaultShortCircuitParameters();
             insertDuplicatedStudy(basicStudyInfos, sourceStudy, LoadflowService.toEntity(newLoadFlowParameters), ShortCircuitService.toEntity(shortCircuitParameters), userId, clonedNetworkUuid, clonedCaseUuid);
+            reindexStudy(basicStudyInfos.getId());
```
SlimaneAmar, line 278: We have already the studuy entity and the networkUuid
Perhaps create a private method in order to avoid a db query ?

dbraquart: here we have the source StudyEntity, but we need to query to get the duplicated StudyEntity in reindexStudy()

SlimaneAmar: just return the entity in `insertDuplicatedStudy`

dbraquart: yes, I did not see it !  -> changed

SlimaneAmar (approved):

PR #363, opened 2023-03-29
Use loadflow provider specific paramaters

dbraquart:


geofjamg (changes requested):

src/main/java/org/gridsuite/study/server/dto/LoadFlowParametersInfos.java
```diff
@@ -0,0 +1,26 @@
...
+import lombok.Getter;
+import java.util.Map;
+
+/**
+ * @author David Braquart <david.braquart@rte-france.com>
+ */
+@Getter
+@AllArgsConstructor
+@Builder
+public class LoadFlowParametersInfos {
```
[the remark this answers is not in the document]
dbraquart: LoadFlowParametersInfos is sent to loadflow-server when a LF computation is started.
Same with SecurityAnalysisParametersInfos (new DTO) and SensitivityAnalysisInputData (existing DTO).
Each DTO contains "Map<String, String> specificParameters", which is used  by each computation server
to create speficic params extensions.
We pass only String values because the "value" column in the load_flow_specific_parameters table is not typed.

LoadFlowParametersValues is the DTO sent to the Front on the Get operation /studies/{studyUuid}/loadflow/parameters.
Because this call is made only a single time when we open the Study (existing design, cf hook useParametersBackend),
we return all parameters, for all LF providers.

src/main/java/org/gridsuite/study/server/dto/SecurityAnalysisParametersInfos.java
```diff
@@ -0,0 +1,27 @@
...
+ * @author David Braquart <david.braquart@rte-france.com>
+ */
+@Getter
+@AllArgsConstructor
+@Builder
+public class SecurityAnalysisParametersInfos {
+
+    private SecurityAnalysisParameters parameters;
+
+    private Map<String, String> loadFlowSpecificParameters;
```
FranckLecuyer, line 26: why not using : List\<LoadFlowSpecificParameterInfos\> ?

src/main/java/org/gridsuite/study/server/dto/SensitivityAnalysisInputData.java
```diff
@@ -126,4 +127,7 @@ public static class SensitivityNodes {
 
     @Schema(description = "Sensitivity parameters")
     private SensitivityAnalysisParameters parameters;
+
+    @Schema(description = "Loadflow model-specific parameters")
+    private Map<String, String> loadFlowSpecificParameters;
```
FranckLecuyer, line 132: why not using : List\<LoadFlowSpecificParameterInfos\> ?

src/main/java/org/gridsuite/study/server/dto/LoadFlowParametersValues.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+@Getter
+@AllArgsConstructor
+@Builder
+@NoArgsConstructor
+public class LoadFlowParametersValues {
+
+    private LoadFlowParameters commonParameters;
+
+    private Map<String, Map<String, Object>> specificParametersPerProvider;
```
FranckLecuyer, line 28: Why Object and not String ?

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/study/server/dto/SecurityAnalysisParametersInfos.java
dbraquart: Yes I could have used that.
I choose the Map because this is what is used to create the extentions (in the compuation servers), and there is no need the declare/use LoadFlowSpecificParameterInfos in every computation server.

Do you think it is better/clearer to use List<LoadFlowSpecificParameterInfos> ?

src/main/java/org/gridsuite/study/server/dto/LoadFlowParametersInfos.java
```diff
@@ -0,0 +1,26 @@
...
+
+    private final LoadFlowParameters commonParameters;
+
+    private final Map<String, String> specificParameters;
```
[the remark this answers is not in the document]
dbraquart: same

src/main/java/org/gridsuite/study/server/dto/SecurityAnalysisParametersInfos.java
FranckLecuyer: No, the change is not needed ...

src/main/java/org/gridsuite/study/server/dto/LoadFlowParametersValues.java
FranckLecuyer: ok

FranckLecuyer (approved):

etiennehomer (approved):

PR #364, opened 2023-03-30
Delete previous short circuit results before running a new computation

FranckLecuyer:


sBouzols (approved):
Code Review OK
Tests OK
(perf problem will be fix in another PR)

PR #385, opened 2023-06-02
Add voltage init parameters

Meklo:


etiennehomer (approved):
Tests OK, code OK

The voltage limits could have been stored in the config server. To be done later maybe if we link this config to the run button like sensi and security analysis

PR #391, opened 2023-06-12
Upgrade to java17

jonenst:


AbdelHedhili (changes requested):
you forgot to set the jdk 17 in github CI

AbdelHedhili (approved):

PR #393, opened 2023-06-14
Adapt duplicate method to allow subtree copy across studies

Meklo:


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1402,7 +1402,7 @@ public void deleteNode(UUID studyUuid, UUID nodeId, boolean deleteChildren, Stri
         startTime.set(System.nanoTime());
         DeleteNodeInfos deleteNodeInfos = new DeleteNodeInfos();
         deleteNodeInfos.setNetworkUuid(networkStoreService.doGetNetworkUuid(studyUuid));
-        boolean invalidateChildrenBuild = !EMPTY_ARRAY.equals(networkModificationTreeService.getNetworkModifications(nodeId));
+        boolean invalidateChildrenBuild = !deleteChildren && !EMPTY_ARRAY.equals(networkModificationTreeService.getNetworkModifications(nodeId));
```
Meklo, line 1405: Fixes an unrelated bug : if a node's children are also deleted it is unnecessary to invalidate them

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -217,17 +217,18 @@ public ResponseEntity<Void> cutAndPasteNodeSubtree(@PathVariable("studyUuid") UU
...
-    @PostMapping(value = "/studies/{studyUuid}/tree/subtrees", params = {"subtreeToCopyParentNodeUuid", "referenceNodeUuid"})
+    @PostMapping(value = "/studies/{targetStudyUuid}/tree/subtrees", params = {"subtreeToCopyParentNodeUuid", "referenceNodeUuid"})
     @Operation(summary = "duplicate a subtree")
     @ApiResponses(value = {
         @ApiResponse(responseCode = "200", description = "The subtree was successfully created"),
         @ApiResponse(responseCode = "403", description = "The subtree can't be copied above the root node nor around itself"),
         @ApiResponse(responseCode = "404", description = "The source study or subtree doesn't exist")})
-    public ResponseEntity<Void> duplicateSubtree(@PathVariable("studyUuid") UUID studyUuid,
+    public ResponseEntity<Void> duplicateSubtree(@Parameter(description = "The study where we want to copy the node") @PathVariable("targetStudyUuid") UUID targetStudyUuid,
+                                                 @Parameter(description = "The copied node original study") @RequestParam(value = "sourceStudyUuid", required = false) UUID sourceStudyUuid,
```
sBouzols, line 227: I think it's better to make both parameter required and call this endpoint with the same value in both parameters when copying subtree in the same study
(`targetStudyUuid` and `sourceStudyUuid`)

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -217,17 +217,18 @@ public ResponseEntity<Void> cutAndPasteNodeSubtree(@PathVariable("studyUuid") UU
...
 
-    @PostMapping(value = "/studies/{studyUuid}/tree/subtrees", params = {"subtreeToCopyParentNodeUuid", "referenceNodeUuid"})
+    @PostMapping(value = "/studies/{targetStudyUuid}/tree/subtrees", params = {"subtreeToCopyParentNodeUuid", "referenceNodeUuid"})
     @Operation(summary = "duplicate a subtree")
     @ApiResponses(value = {
         @ApiResponse(responseCode = "200", description = "The subtree was successfully created"),
         @ApiResponse(responseCode = "403", description = "The subtree can't be copied above the root node nor around itself"),
         @ApiResponse(responseCode = "404", description = "The source study or subtree doesn't exist")})
-    public ResponseEntity<Void> duplicateSubtree(@PathVariable("studyUuid") UUID studyUuid,
+    public ResponseEntity<Void> duplicateSubtree(@Parameter(description = "The study where we want to copy the node") @PathVariable("targetStudyUuid") UUID targetStudyUuid,
```
sBouzols, line 226: MAybe we should keep `{studyUuid}` in the REST path for a resource and just call the variable here with the target prefix
`@PathVariable("studyUuid") UUID targetStudyUuid,`
should be good.

src/main/java/org/gridsuite/study/server/StudyController.java
Meklo: Done, i thus removed the ternary operation when calling the service method

src/main/java/org/gridsuite/study/server/StudyController.java
Meklo: Done, everything still works as intended

jonenst (approved):

sBouzols (approved):
Code Review OK
Test OK

PR #397, opened 2023-06-15
Prevent node children invalidation in case they are also deleted

Meklo:


sBouzols (approved):
Code Review OK
Test OK

PR #418, opened 2023-08-10
Dynamic simulation - Event configuration

thangqp:
- This back-end PR provides REST CRUD Endpoints for events that are used in the dynamic simulation.

- Events are associated to nodes, at this moment no inheritance for events is implemented between nodes 

**Front-end PR**: https://github.com/gridsuite/gridstudy-app/pull/1429 (to see Demo)
**Back-end PR**: https://github.com/gridsuite/dynamic-simulation-server/pull/65


jonenst (changes requested):
get user from the controller explicitly like our other code paths for consistency and remove auditaware. Otherwise good for me

thangqp:
> get user from the controller explicitly like our other code paths for consistency and remove auditaware. Otherwise good for me

@Jon, I have removed the solution which relies on Interface AuditorAware but I can not propagate correctly the userId to entities. In fact, the audit is better managed if we set the audit info just before INSERT or UPDATE to DB because Hibernate uses a dirty detection mechanism after calling repository.save() a transient entity.

Could you see my solution in the commit that use @PrePersist and @PreUpdate for Embedable Audit class, pls: https://github.com/gridsuite/study-server/pull/418/commits/81d69da0a59d07bfe34cb59668c99a3949647d92

jonenst (approved):

PR #428, opened 2023-08-31
Add dc power factor when getting element(s) infos from map-server

FranckLecuyer:


dbraquart (approved):
code: ok
tests: ok

Walid-Sahnoun (approved):
code ok
test ok

PR #433, opened 2023-09-08
Restore nodes

AbdelHedhili:


src/main/resources/db/changelog/changesets/changelog_20230825T115511Z.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+                   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                   xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+    <changeSet author="hediliabd" id="1252713657904-1">
+        <addColumn tableName="node">
+            <column name="stashed" type="boolean"/>
```
Meklo, line 7: You could default it to false so that older studies won't break, I got an error when deleting a preexisting node because stashed was null

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1604,6 +1602,37 @@ public void deleteNode(UUID studyUuid, UUID nodeId, boolean deleteChildren, Stri
         notificationService.emitElementUpdated(studyUuid, userId);
     }
 
+    @Transactional
+    public void stashNode(UUID studyUuid, UUID nodeId, boolean stashChildren, String userId) {
+        AtomicReference<Long> startTime = new AtomicReference<>(null);
+        startTime.set(System.nanoTime());
+        DeleteNodeInfos deleteNodeInfos = new DeleteNodeInfos();
```
Meklo, line 1609: I don't see how DeleteNodeInfos is needed for this use case, if i'm not mistaken it's used during a node's deletion to keep track of its modifications, network, and various results after it's been deleted from the db but in case of a stash is it really useful ?

src/main/resources/db/changelog/changesets/changelog_20230825T115511Z.xml
AbdelHedhili: indeed, added

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -1126,16 +1127,48 @@ public ResponseEntity<AbstractNode> createNode(@RequestBody AbstractNode node,
...
+    @GetMapping(value = "/studies/{studyUuid}/tree/nodes/stash")
+    @Operation(summary = "Get the list of nodes in the trash for a given study")
+    @ApiResponses(value = {
+            @ApiResponse(responseCode = "200", description = "the list of nodes in the trash")})
+    public ResponseEntity<List<Pair<AbstractNode, Integer>>> getStashedNodes(@Parameter(description = "study uuid") @PathVariable("studyUuid") UUID studyUuid) {
+        return ResponseEntity.ok().body(studyService.getStashedNodes(studyUuid));
+    }
+
+    @PostMapping(value = "/studies/{studyUuid}/tree/nodes/{nodeId}/restore")
+    @Operation(summary = "restore node on below the given anchor node")
```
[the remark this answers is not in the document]
AbdelHedhili: removed the "on"

Meklo (approved):

souissimai (approved):

PR #437, opened 2023-09-19
Supervision Add endpoint to remove all indexed-equipments and tombstoned indexed equipments from ES

sBouzols:


src/main/java/org/gridsuite/study/server/service/SupervisionService.java
```diff
@@ -73,6 +87,24 @@ public Integer deleteComputationResults(ComputationType computationType, boolean
         }
     }
 
+    private void updateStudiesIndexationStatus(StudyIndexationStatus indexationStatus) {
+        List<StudyEntity> studies = studyRepository.findAll();
```
jonenst, line 91: Did you think about findAll and foreach performance for big databases ? I think we are kind of near the limit here, for studies we expect at maximum 100k I guess, so it could work (and especially for smaller databases with 1k or 10k studies it will work). Maybe add a comment explaining the reasoning in the code

jonenst: The alternative is to run this in an asynchronous method

Meklo (approved):

src/main/java/org/gridsuite/study/server/service/SupervisionService.java
sBouzols: I thought to update all the column in the study table directly but it is recommended to pass throught entities.
Then I thought those updates could be batched but batch_size and all aren't activated in study-server.
Finally I thought that for around maybe ~100 studies today, it's ok this way.

jonenst (approved):

PR #446, opened 2023-10-02
set inUpstreamBuiltParentNode to false by default for get equipement …

jamal-khey:
set all end points that use inUpstreamBuiltParentNode=true as a default value to'true'

### DIFF
--- a/src/main/java/org/gridsuite/study/server/StudyController.java
+++ b/src/main/java/org/gridsuite/study/server/StudyController.java
@@ -365,7 +365,7 @@ public class StudyController {
             @PathVariable("studyUuid") UUID studyUuid,
             @PathVariable("nodeUuid") UUID nodeUuid,
             @PathVariable("voltageLevelId") String voltageLevelId,
-            @Parameter(description = "Should get in upstream built node ?") @RequestParam(value = "inUpstreamBuiltParentNode", required = false, defaultValue = "true") boolean inUpstreamBuiltParentNode) {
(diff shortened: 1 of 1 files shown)


SlimaneAmar (approved):

FranckLecuyer (approved):

antoinebhs (changes requested):
There are some API calls that rely on the inUpstreamBuildParentNode set to true. We should check carefully each call and decide whether or not inUpstreamBuildParentNode should be set to true. 
With the changes, I have an internal error 500 when I try to delete an HVDC line on a node that is not built. I have no issues when I'm on the main branch. See Antoine/test/recollement on dev-OPF as an example. I think there are other calls that might be problematic. I only tested https://github.com/gridsuite/gridstudy-app/blob/c77dadb572feb8d092981851de3fc4b001498226/src/services/study/network-map.js#L22.

jamal-khey:
> There are some API calls that rely on the inUpstreamBuildParentNode set to true. We should check carefully each call and decide whether or not inUpstreamBuildParentNode should be set to true. With the changes, I have an internal error 500 when I try to delete an HVDC line on a node that is not built. I have no issues when I'm on the main branch. See Antoine/test/recollement on dev-OPF as an example. I think there are other calls that might be problematic. I only tested https://github.com/gridsuite/gridstudy-app/blob/c77dadb572feb8d092981851de3fc4b001498226/src/services/study/network-map.js#L22.

after a discution with Christelle , it has been decided to keep this parameter to true for the end point.

antoinebhs (approved):

PR #455, opened 2023-10-09
logs on demand

dbraquart:
This PR works only with branch _logs_optimization_ from _report-server_ (no PR at this time)

### DIFF
--- a/src/test/java/org/gridsuite/study/server/service/ReportServiceTest.java
+++ b/src/test/java/org/gridsuite/study/server/service/ReportServiceTest.java
@@ -173,7 +173,7 @@ public class ReportServiceTest {
         RootNode rootNode = createRoot();
         ReporterModel expectedRootReporter = getNodeReport(ROOT_NODE_REPORT_UUID.toString(), rootNode.getId().toString());
 
-        MvcResult mvcResult = mockMvc.perform(get("/v1/studies/{studyUuid}/nodes/{nodeUuid}/report?nodeOnlyReport=true", rootNode.getStudyId(), rootNode.getId()))
+        MvcResult mvcResult = mockMvc.perform(get("/v1/studies/{studyUuid}/nodes/{nodeUuid}/parent-nodes-report?nodeOnlyReport=true", rootNode.getStudyId(), rootNode.getId()))
             .andExpectAll(status().isOk(), content().contentType(MediaType.APPLICATION_JSON))
             .andReturn();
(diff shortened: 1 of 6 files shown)


src/test/java/org/gridsuite/study/server/service/ReportServiceTest.java
```diff
@@ -129,10 +128,15 @@ public Object getTypeId() {
...
             public MockResponse dispatch(RecordedRequest request) {
                 String path = Objects.requireNonNull(request.getPath());
-                if (path.matches("/v1/reports/.*")) {
+                if (path.matches("/v1/reports/.*\\?defaultName=.*")) {
                     return new MockResponse().setResponseCode(HttpStatus.OK.value())
-                        .setBody(mapper.writeValueAsString(getNodeReport(Objects.requireNonNull(request.getRequestUrl()).pathSegments().get(2), request.getRequestUrl().queryParameter(QUERY_PARAM_REPORT_DEFAULT_NAME)).getSubReporters()))
-                        .addHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE);
+                            .setBody(mapper.writeValueAsString(getNodeReport(Objects.requireNonNull(request.getRequestUrl()).pathSegments().get(2), request.getRequestUrl().queryParameter(QUERY_PARAM_REPORT_DEFAULT_NAME)).getSubReporters()))
+                            .addHeader(HttpHeaders.CONTENT_TYPE, MediaType.APPLICATION_JSON_VALUE);
+                } else if (path.matches("/v1/subreports/.*\\?severityLevels=.*")) {
```
EstherDarkish, line 135: Is it worth it to include `&` in these patterns, in case we have something like `/v1/reports/1234?hello=true&defaultName` ?

dbraquart: I dont get this comment

SlimaneAmar (approved):

src/test/java/org/gridsuite/study/server/service/ReportServiceTest.java
EstherDarkish: The current regex will only match if severityLevels is the first parameter in the URL. If it is the second, then it would match "&severityLevel=" instead of "?severityLevel="

dbraquart: ok, I get it.
But we have only one param currently, so severityLevels is always first.

PR #457, opened 2023-10-11
Add pagination and filtering for one bus short circuit analysis

flomillot:
Update the endpoint for paginated fault results to be able to get also feeder results.


src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -642,17 +642,22 @@ public ResponseEntity<String> getShortCircuitResult(@Parameter(description = "st
...
         @ApiResponse(responseCode = "404", description = "The short circuit analysis has not been found")})
-    public ResponseEntity<String> getShortCircuitAnalysisFaultResultsPage(@Parameter(description = "study UUID") @PathVariable("studyUuid") UUID studyUuid,
-                                                               @Parameter(description = "nodeUuid") @PathVariable("nodeUuid") UUID nodeUuid,
-                                                               @Parameter(description = "Full or only those with limit violations or none fault results") @RequestParam(name = "mode", required = false, defaultValue = "WITH_LIMIT_VIOLATIONS") String mode,
-                                                               Pageable pageable) {
-        String faultResultsPage = shortCircuitService.getShortCircuitAnalysisFaultResultsPage(nodeUuid, mode, pageable);
-        return faultResultsPage != null ? ResponseEntity.ok().body(faultResultsPage) :
+    public ResponseEntity<String> getShortCircuitAnalysisResultsPage(@Parameter(description = "study UUID") @PathVariable("studyUuid") UUID studyUuid,
```
flomillot, line 650: I'm not sure about this, should I make 2 endpoints ?

src/main/java/org/gridsuite/study/server/service/shortcircuit/ShortCircuitService.java
```diff
@@ -116,51 +117,70 @@ private String getShortCircuitAnalysisResultResourcePath(UUID nodeUuid, Shortcir
...
+                .queryParam("size", pageable.getPageSize());
+
+        if (filters != null && !filters.isEmpty()) {
+            builder.queryParam("filters", filters);
         }
-        String faultResultsPath = getShortCircuitAnalysisFaultResultsResourcePath(nodeUuid);
-        if (faultResultsPath == null) {
-            return null;
+
+        for (Sort.Order order : pageable.getSort()) {
```
sBouzols, line 162: We should add a function which take a `Pageable` and return query params because it will be used on other places where pagination is needed

src/main/java/org/gridsuite/study/server/StudyController.java
sBouzols: I would have made 2 endpoints yes, one for faultResultPage and one for FeederResultPage yes.

flomillot: It have been decided to remove this endpoint and use only the legacy one, as for sensi

src/main/java/org/gridsuite/study/server/service/shortcircuit/ShortCircuitService.java
flomillot: Good idea, done

sBouzols (approved):
Code Review OK
Tests Done. Some minor problems but will be fixed on the next part for all buses results

anistouri (approved):

PR #469, opened 2023-11-10
Fix: stashed modifications not excluded during node building

sBouzols:
`stashNetworkModifications` must `addModificationsToExclude` instead of `removeModificationsToExclude`
Added some TU to check this

### DIFF
--- a/src/main/java/org/gridsuite/study/server/notification/NotificationService.java
+++ b/src/main/java/org/gridsuite/study/server/notification/NotificationService.java
@@ -74,6 +74,7 @@ public class NotificationService {
 
     public static final String MODIFICATIONS_CREATING_IN_PROGRESS = "creatingInProgress";
     public static final String MODIFICATIONS_DELETING_IN_PROGRESS = "deletingInProgress";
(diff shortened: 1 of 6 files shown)


src/main/java/org/gridsuite/study/server/notification/NotificationService.java
```diff
@@ -74,6 +74,7 @@ public class NotificationService {
 
     public static final String MODIFICATIONS_CREATING_IN_PROGRESS = "creatingInProgress";
     public static final String MODIFICATIONS_DELETING_IN_PROGRESS = "deletingInProgress";
+    public static final String MODIFICATIONS_RESTORING_IN_PROGRESS = "restoringInProgress";
```
jamal-khey, line 77: cette notification est a traiter coté front ?
je crois qu'il y a quelque chose a faire coté src/components/graph/menus/network-modification-node-editor.js:433

sBouzols: True !
Added here : https://github.com/gridsuite/gridstudy-app/pull/1685

jamal-khey (approved):

PR #473, opened 2023-11-14
delete unused endpoint (voltage-levelsequipments)

achour94:


jamal-khey (approved):
only one remark, the PR title do not match the content
should probably be 'rename voltage-level-equipments endpoint'  not delete unused endpoint

souissimai (approved):

PR #475, opened 2023-11-15
add actual and upcoming overload for limiViolations

souissimai:


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1123,10 +1123,10 @@ public Integer getContingencyCount(UUID studyUuid, List<String> contingencyListN
         return actionsService.getContingencyCount(networkuuid, variantId, contingencyListNames);
     }
 
+    // TO DELETE
```
FranckLecuyer, line 1126: So, remove this method, which is not used anymore

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/study/server/service/StudyService.java
souissimai: done

dbraquart (changes requested):
code: a rename requested, if possible

FranckLecuyer (approved):

src/main/java/org/gridsuite/study/server/service/StudyService.java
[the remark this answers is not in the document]
souissimai: done

dbraquart (approved):
code: ok
tests: ok

PR #496, opened 2023-12-18
Replace post http method by get method for sensi factors count

SlimaneAmar:


TheMaskedTurtle (approved):

TheMaskedTurtle (commented):
Same remark as for study sensitivity-analysis-server as it is just a forwarded request, why not 3 explicit lists ?

PR #499, opened 2023-12-21
refacto computations handling

Mathieu-Deharbe:
Refactorisation of the computation handling in the Consumer service + adds a toString to the ComputationType enum.

This is only a refacto so when I found something suspect I only added a comment. Those will be handled in a following US tied to this one.

### DIFF
--- a/src/main/java/org/gridsuite/study/server/dto/ComputationType.java
+++ b/src/main/java/org/gridsuite/study/server/dto/ComputationType.java
@@ -6,7 +6,7 @@ public enum ComputationType {
     SENSITIVITY_ANALYSIS("SensitivityAnalysis", "Sensitivity analysis"),
     SHORT_CIRCUIT("ShortCircuitAnalysis", "Short circuit analysis"),
     VOLTAGE_INITIALIZATION("", "Voltage init"),
-    DYNAMIC_SIMULATION("","Dynamic simulation");
+    DYNAMIC_SIMULATION("", "Dynamic simulation");
 
     public final String subReporterKey;
     public final String label;
--- a/src/main/java/org/gridsuite/study/server/dto/ComputationType.java
+++ b/src/main/java/org/gridsuite/study/server/dto/ComputationType.java
@@ -1,10 +1,23 @@
 package org.gridsuite.study.server.dto;
 
 public enum ComputationType {
-    LOAD_FLOW,
-    SECURITY_ANALYSIS,
-    SENSITIVITY_ANALYSIS,
-    SHORT_CIRCUIT,
-    VOLTAGE_INITIALIZATION,
-    DYNAMIC_SIMULATION
+    LOAD_FLOW("LoadFlow", "Loadflow"),
+    SECURITY_ANALYSIS("SecurityAnalysis", "Security analysis"),
+    SENSITIVITY_ANALYSIS("SensitivityAnalysis", "Sensitivity analysis"),
+    SHORT_CIRCUIT("ShortCircuitAnalysis", "Short circuit analysis"),
+    VOLTAGE_INITIALIZATION("", "Voltage init"),
+    DYNAMIC_SIMULATION("","Dynamic simulation");
+
+    public final String subReporterKey;
+    public final String label;
+
+    ComputationType(String subReporterKey, String label) {
+        this.subReporterKey = subReporterKey;
+        this.label = label;
+    }
+
+    @Override
+    public String toString() {
+        return this.label;
+    }
 }
--- a/src/main/java/org/gridsuite/study/server/dto/ComputationType.java
+++ b/src/main/java/org/gridsuite/study/server/dto/ComputationType.java
@@ -16,8 +16,7 @@ public enum ComputationType {
         this.label = label;
     }
 
-    @Override
-    public String toString() {
+    public String getLabel() {
         return this.label;
     }
 }
--- a/src/main/java/org/gridsuite/study/server/dto/ComputationType.java
+++ b/src/main/java/org/gridsuite/study/server/dto/ComputationType.java
@@ -2,59 +2,42 @@ package org.gridsuite.study.server.dto;
 
 import lombok.Getter;
 import org.gridsuite.study.server.notification.NotificationService;
-import org.gridsuite.study.server.service.shortcircuit.ShortcircuitAnalysisType;
-
-import static org.gridsuite.study.server.service.shortcircuit.ShortcircuitAnalysisType.ONE_BUS;
 
 @Getter
 public enum ComputationType {
-    LOAD_FLOW("LoadFlow", "Load flow", "loadFlowResultUuid",
+    LOAD_FLOW("Load flow", "loadFlowResultUuid",
             NotificationService.UPDATE_TYPE_LOADFLOW_STATUS, NotificationService.UPDATE_TYPE_LOADFLOW_RESULT,
             NotificationService.UPDATE_TYPE_LOADFLOW_FAILED),
-    SECURITY_ANALYSIS("SecurityAnalysis", "Security analysis", "securityAnalysisResultUuid",
+    SECURITY_ANALYSIS("Security analysis", "securityAnalysisResultUuid",
             NotificationService.UPDATE_TYPE_SECURITY_ANALYSIS_STATUS, NotificationService.UPDATE_TYPE_SECURITY_ANALYSIS_RESULT,
             NotificationService.UPDATE_TYPE_SECURITY_ANALYSIS_FAILED),
-    SENSITIVITY_ANALYSIS("SensitivityAnalysis", "Sensitivity analysis", "sensitivityAnalysisResultUuid",
+    SENSITIVITY_ANALYSIS("Sensitivity analysis", "sensitivityAnalysisResultUuid",
             NotificationService.UPDATE_TYPE_SENSITIVITY_ANALYSIS_STATUS, NotificationService.UPDATE_TYPE_SENSITIVITY_ANALYSIS_RESULT,
             NotificationService.UPDATE_TYPE_SENSITIVITY_ANALYSIS_FAILED),
-    SHORT_CIRCUIT("ShortCircuitAnalysis", "Short circuit analysis", "shortCircuitAnalysisResultUuid",
-            NotificationService.UPDATE_TYPE_ONE_BUS_SHORT_CIRCUIT_STATUS, NotificationService.UPDATE_TYPE_ONE_BUS_SHORT_CIRCUIT_RESULT,
-            NotificationService.UPDATE_TYPE_ONE_BUS_SHORT_CIRCUIT_FAILED, ONE_BUS),
-    VOLTAGE_INITIALIZATION("", "Voltage init", "voltageInitResultUuid",
+    SHORT_CIRCUIT("Short circuit analysis", "shortCircuitAnalysisResultUuid",
+            NotificationService.UPDATE_TYPE_SHORT_CIRCUIT_STATUS, NotificationService.UPDATE_TYPE_SHORT_CIRCUIT_RESULT,
+            NotificationService.UPDATE_TYPE_SHORT_CIRCUIT_FAILED),
+    VOLTAGE_INITIALIZATION("Voltage init", "voltageInitResultUuid",
             NotificationService.UPDATE_TYPE_VOLTAGE_INIT_STATUS, NotificationService.UPDATE_TYPE_VOLTAGE_INIT_RESULT,
             NotificationService.UPDATE_TYPE_VOLTAGE_INIT_FAILED),
-    DYNAMIC_SIMULATION("", "Dynamic simulation", "dynamicSimulationResultUuid",
+    DYNAMIC_SIMULATION("Dynamic simulation", "dynamicSimulationResultUuid",
             NotificationService.UPDATE_TYPE_DYNAMIC_SIMULATION_STATUS, NotificationService.UPDATE_TYPE_DYNAMIC_SIMULATION_RESULT,
-            NotificationService.UPDATE_TYPE_DYNAMIC_SIMULATION_FAILED);
+            NotificationService.UPDATE_TYPE_DYNAMIC_SIMULATION_FAILED),
+    SHORT_CIRCUIT_ONE_BUS("One bus Short circuit analysis", "oneBusShortCircuitAnalysisResultUuid",
+            NotificationService.UPDATE_TYPE_ONE_BUS_SHORT_CIRCUIT_STATUS, NotificationService.UPDATE_TYPE_ONE_BUS_SHORT_CIRCUIT_RESULT,
+            NotificationService.UPDATE_TYPE_ONE_BUS_SHORT_CIRCUIT_FAILED);
 
-    public final String subReporterKey;
     private final String label; // used for logs
-    private String resultUuidLabel;
-    private ShortcircuitAnalysisType oneBus; // only useful for SHORT_CIRCUIT
-    private String updateStatusType;
-    private String updateResultType;
-    private String updateFailedType;
+    private final String resultUuidLabel;
+    private final String updateStatusType;
+    private final String updateResultType;
+    private final String updateFailedType;
 
-    public void setBusMode(ShortcircuitAnalysisType oneBus) {
-        this.oneBus = oneBus;
-        this.resultUuidLabel = this.oneBus == ONE_BUS ? "oneBusShortCircuitAnalysisResultUuid" : "shortCircuitAnalysisResultUuid";
-        this.updateStatusType = this.oneBus == ONE_BUS ? NotificationService.UPDATE_TYPE_ONE_BUS_SHORT_CIRCUIT_STATUS : NotificationService.UPDATE_TYPE_SHORT_CIRCUIT_STATUS;
-        this.updateResultType = this.oneBus == ONE_BUS ? NotificationService.UPDATE_TYPE_ONE_BUS_SHORT_CIRCUIT_RESULT : NotificationService.UPDATE_TYPE_SHORT_CIRCUIT_RESULT;
-        this.updateFailedType = this.oneBus == ONE_BUS ? NotificationService.UPDATE_TYPE_ONE_BUS_SHORT_CIRCUIT_FAILED : NotificationService.UPDATE_TYPE_SHORT_CIRCUIT_FAILED;
-    }
-
-    ComputationType(String subReporterKey, String label, String resultUuidLabel, String updateStatusType, String updateResultType, String updateFailedType) {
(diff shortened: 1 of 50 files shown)


src/main/java/org/gridsuite/study/server/dto/ComputationType.java
```diff
@@ -1,16 +1,22 @@
 package org.gridsuite.study.server.dto;
 
 public enum ComputationType {
-    LOAD_FLOW("LoadFlow"),
-    SECURITY_ANALYSIS("SecurityAnalysis"),
-    SENSITIVITY_ANALYSIS("SensitivityAnalysis"),
-    SHORT_CIRCUIT("ShortCircuitAnalysis"),
-    VOLTAGE_INITIALIZATION(""),
-    DYNAMIC_SIMULATION("");
+    LOAD_FLOW("LoadFlow", "Loadflow"),
```
jamal-khey, line 4: a space maybe in "Load flow"

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -398,367 +238,287 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
-                } catch (JsonProcessingException e) {
-                    LOGGER.error(e.toString());
+    public void consumeCalculationResult(Message<String> msg, ComputationType computationType) {
+        UUID resultUuid = null;
+        String resultId = msg.getHeaders().get(RESULT_UUID, String.class);
+        if (resultId != null) {
+            resultUuid = UUID.fromString(resultId);
+        }
+        String receiver = msg.getHeaders().get(HEADER_RECEIVER, String.class);
+        if (receiver != null) {
```
jamal-khey, line 368: the test done before is '!Strings.isBlank(receiver)'  that checks if a String is whitespace, empty ("") or null. do you have a reasons for changing it ?
```suggestion
        !Strings.isBlank(receiver)
```

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -398,367 +238,287 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
 
-                    // send notification for stopped computation
-                    UUID studyUuid = networkModificationTreeService.getStudyUuidForNodeId(receiverObj.getNodeUuid());
-                    notificationService.emitStudyChanged(studyUuid, receiverObj.getNodeUuid(), NotificationService.UPDATE_TYPE_SENSITIVITY_ANALYSIS_STATUS);
-                } catch (JsonProcessingException e) {
-                    LOGGER.error(e.toString());
+    public void consumeCalculationResult(Message<String> msg, ComputationType computationType) {
+        UUID resultUuid = null;
+        String resultId = msg.getHeaders().get(RESULT_UUID, String.class);
+        if (resultId != null) {
```
jamal-khey, line 364: resultId can be null ? i don't see this test in the old  code.

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -398,367 +238,287 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
+                    case SECURITY_ANALYSIS: {
+                        networkModificationTreeService.updateSecurityAnalysisResultUuid(receiverObj.getNodeUuid(), null);
+                        updateType = NotificationService.UPDATE_TYPE_SECURITY_ANALYSIS_STATUS;
+                    } break;
+                    case SENSITIVITY_ANALYSIS: {
+                        networkModificationTreeService.updateSensitivityAnalysisResultUuid(receiverObj.getNodeUuid(), null);
+                        updateType = NotificationService.UPDATE_TYPE_SENSITIVITY_ANALYSIS_STATUS;
+                    } break;
+                    case SHORT_CIRCUIT: {
+                        // TODO : this Stopped function is the only one when SHORT_CIRCUIT doesn't handle buses => might cause a bug
```
jamal-khey, line 338: this todo is something you have discovred ?

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -398,367 +238,287 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
-
-                    LOGGER.info("Sensitivity analysis stopped for node '{}'", receiverObj.getNodeUuid());
+    public void consumeCalculationStopped(Message<String> msg, ComputationType computationType) {
+        String receiver = msg.getHeaders().get(HEADER_RECEIVER, String.class);
+        if (receiver != null) {
+            NodeReceiver receiverObj;
+            try {
+                receiverObj = objectMapper.readValue(URLDecoder.decode(receiver, StandardCharsets.UTF_8), NodeReceiver.class);
+
+                LOGGER.info("{} stopped for node '{}'", computationType.getLabel(), receiverObj.getNodeUuid());
```
jamal-khey, line 320: may be this comment should be after the switch case ? or change it to '{} is stopping ...'  because the service is not stopped yet

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
Mathieu-Deharbe: Yes I find it very suspicious but I am supposed to only refactor so I added a TODO.

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
sBouzols: it's a channel consumption in study-server of the stopped message from the rabbitMQ queue then the calculation is already stopped in the dedicated calculation server.
But maybe it's a good idea to move this log or add one at the end of this try scope to trace that the resultUuid for this type of calculation is reset.

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
sBouzols: It's not possibe to stop a ONE_BUS shortcircuit for now, because it's not done with the button like other calculation. By the way, here in the server java code, nothing disallow to manage it properly now if we have HEADER_BUS_ID in the Header... but do we have it ?

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -398,367 +238,287 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
+                        updateResultType = NotificationService.UPDATE_TYPE_VOLTAGE_INIT_RESULT;
+                    } break;
+                    case DYNAMIC_SIMULATION: {
+                        networkModificationTreeService.updateDynamicSimulationResultUuid(receiverObj.getNodeUuid(), resultUuid);
+                        updateStatusType = NotificationService.UPDATE_TYPE_DYNAMIC_SIMULATION_STATUS;
+                        updateResultType = NotificationService.UPDATE_TYPE_DYNAMIC_SIMULATION_RESULT;
+                    } break;
                 }
+
+                UUID studyUuid = networkModificationTreeService.getStudyUuidForNodeId(receiverObj.getNodeUuid());
```
sBouzols, line 425: We don't have the studyUuid in the message headers ? that's a pity to make a BDD query for this no ?
Arf, no, Calcultion services don't know about studyUuid ok then.

Mathieu-Deharbe: Indeed. I don't really the whole flux and am only supposed to refactor, so for now, I can only assume that there is a good reason.

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
Mathieu-Deharbe: Sometimes there is just a `!= null` test and sometimes `!Strings.isBlank(receiver)`. I tried to follow the same pattern when refactoring and sometimes kept both but indeed it doesn't make much sense. I will just use isBlank in both cases.

src/main/java/org/gridsuite/study/server/dto/ComputationType.java
Mathieu-Deharbe: ok

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
Mathieu-Deharbe: Nope you are right : HEADER_BUS_ID is only sent in case of success or failure. I remove the TODO.

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
Mathieu-Deharbe: Agreed, moved.

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
Mathieu-Deharbe: It was only in one of the code version that I reformated. I kept it but it looks very unlikely.

src/main/java/org/gridsuite/study/server/dto/ComputationType.java
```diff
@@ -1,16 +1,60 @@
...
 
     public final String subReporterKey;
+    private final String label; // used for logs
+    private String resultUuidLabel;
+    private ShortcircuitAnalysisType oneBus; // only useful for SHORT_CIRCUIT
+    private String updateStatusType;
+    private String updateResultType;
+    private String updateFailedType;
+
+    public void setBusMode(ShortcircuitAnalysisType oneBus) {
```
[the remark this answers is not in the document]
Mathieu-Deharbe: Yep and it was not a good idea to update an enum value. 

I added one more value to the enum (SHORT_CIRCUIT_ONE_BUS). Not perfect because it turns an enum value into two but everyhing is much clearer.

What do you think ? How would you write your subtype suggestion ?

src/main/java/org/gridsuite/study/server/service/SupervisionService.java
```diff
@@ -90,6 +90,7 @@ public Integer deleteComputationResults(ComputationType computationType, boolean
             case SENSITIVITY_ANALYSIS:
                 return dryRun ? sensitivityAnalysisService.getSensitivityAnalysisResultsCount() : deleteSensitivityAnalysisResults();
             case SHORT_CIRCUIT:
+            case SHORT_CIRCUIT_ONE_BUS:
```
dbraquart, line 93: admin-tools is currently passing SHORT_CIRCUIT to remove all results from short-circuit repository (both all-buses and 1-bus results). In deed this repository cannot make the difference.
So I think it's useless to add SHORT_CIRCUIT_ONE_BUS here cause it wont be passed by admin-tool.

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -398,367 +239,199 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
+            try {
+                receiverObj = objectMapper.readValue(URLDecoder.decode(receiver, StandardCharsets.UTF_8), NodeReceiver.class);
+
+                ComputationType updatedComputationType = computationType;
+                if (updatedComputationType == SHORT_CIRCUIT) {
+                    String busId = msg.getHeaders().get(HEADER_BUS_ID, String.class);
+                    if (!StringUtils.isEmpty(busId)) {
+                        updatedComputationType = SHORT_CIRCUIT_ONE_BUS;
+                    }
+                }
```
sBouzols, line 268: maybe we could move this code block into `consumeShortCircuitAnalysisFailed` then we don't pollute generic code with some really shortcircuit specific code ?

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -398,367 +239,199 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
```
sBouzols, line 328: same and move it in consumeShortCircuitAnalysisResult

src/main/java/org/gridsuite/study/server/service/SupervisionService.java
sBouzols: We could manage both separately and split `getShortCircuitResultsCount` and `deleteShortcircuitResults` methods or consider that giving SHORT_CIRCUIT or SHORT_CIRCUIT_ONE_BUS it delete both results.
Maybe we could add a TODO, but it doesn't change anything here, the same admin-tools script which call with SHORT_CIRCUIT will still get both results count and delete both results type

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -945,11 +951,11 @@ private LoadFlowParametersEntity createParametersEntity(LoadFlowParametersValues
     }
 
     private void deleteLoadFlowResult(UUID studyUuid) {
-        List<UUID> loadFlowResultUuids = networkModificationTreeService.getLoadFlowResultUuids(studyUuid);
+        List<UUID> loadFlowResultUuids = networkModificationTreeService.getComputationResultUuids(studyUuid, LOAD_FLOW);
```
sBouzols, line 954: I don't want this PR to become a nightmare but see this delete*Result could be factorized as well :smile:

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1181,23 +1187,23 @@ public String getNeworkAreaDiagram(UUID studyUuid, UUID nodeUuid, List<String> v
     }
 
     public void invalidateSecurityAnalysisStatusOnAllNodes(UUID studyUuid) {
-        securityAnalysisService.invalidateSaStatus(networkModificationTreeService.getStudySecurityAnalysisResultUuids(studyUuid));
+        securityAnalysisService.invalidateSaStatus(networkModificationTreeService.getComputationResultUuids(studyUuid, SECURITY_ANALYSIS));
```
sBouzols, line 1190: another factorization possibility...

sBouzols: Maybe we could do this in another PR later...

src/main/java/org/gridsuite/study/server/service/StudyService.java
Mathieu-Deharbe: Are you sure ? I only found one 'delete*Result' (this one) and it doesn't seem to be used at all. Technically those in the Services could be  refactored (by creating a parent class for ShortCircuitService etc) but I am not sure that I have the will to start this.

src/main/java/org/gridsuite/study/server/service/StudyService.java
Mathieu-Deharbe: Fine, I will do it :) . And what a surprise : it leads to even more refactoring tied to this function !

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -398,367 +239,199 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
-                NodeReceiver receiverObj;
-                try {
-                    receiverObj = objectMapper.readValue(URLDecoder.decode(receiver, StandardCharsets.UTF_8), NodeReceiver.class);
-
-                    LOGGER.info("Sensitivity analysis result '{}' available for node '{}'", resultUuid, receiverObj.getNodeUuid());
-
-                    // update DB
-                    updateSensitivityAnalysisResultUuid(receiverObj.getNodeUuid(), resultUuid);
+    /**
+     * parse the error message from the computation microservice and uses its data to notify the front
```
jamal-khey, line 243: la fonction ne parse pas vraiment le message
```suggestion
     * Consumes a calculation failed message and notify the front based on the computation type.
```

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -398,367 +239,199 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
-                    LOGGER.info("Sensitivity analysis stopped for node '{}'", receiverObj.getNodeUuid());
+                LOGGER.info("{} result '{}' available for node '{}'",
+                        updatedComputationType.getLabel(),
+                        resultUuid,
+                        receiverObj.getNodeUuid());
 
-                    // delete sensitivity analysis result in database
-                    updateSensitivityAnalysisResultUuid(receiverObj.getNodeUuid(), null);
+                // update DB
+                networkModificationTreeService.updateComputationResultUuid(receiverObj.getNodeUuid(), resultUuid, updatedComputationType);
```
jamal-khey, line 337: i agree with sylvain comment , this code can be moved to networkModificationTreeService to hide the specific code ?

jamal-khey (approved):
Code and tests OK
some comments are still remaining but i consider them as a bonus

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
Mathieu-Deharbe: Agreed. Done.

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
Mathieu-Deharbe: Very good idea. Done.

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
Mathieu-Deharbe: Very good idea. Done.

src/main/java/org/gridsuite/study/server/service/StudyService.java
Mathieu-Deharbe: Changed my mind : All the computations use their own services. I don't think there is much to refactorize here. (unless the various services classes are refactored themselves, inheriting from a new parent abstract class ?)

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
Mathieu-Deharbe: Changed.

src/main/java/org/gridsuite/study/server/service/StudyService.java
sBouzols: Well it seems that this method isn't used anymore then maybe just remove it ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
sBouzols: true it's ok for me.

src/main/java/org/gridsuite/study/server/service/StudyService.java
Mathieu-Deharbe: I dared not but your encouragement decided me ! Done.

PR #501, opened 2023-12-27
Dynamic report severity list

lilian-houdelet:


flomillot (changes requested):
Please update the swagger doc of the endpoints

dbraquart (changes requested):
As discussed, study-server should not be modified for your needs.

PR #512, opened 2024-01-17
Fixed node unbuild on empty modifications trash

ne0ds:


AAJELLAL (approved):
Test and code are ok.

PR #530, opened 2024-02-15
add a rest api to apply a filter on a study root node

dbraquart:


src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -1927,7 +1927,15 @@ public ResponseEntity<String> evaluateFilter(
             @Parameter(description = "Node uuid") @PathVariable("nodeUuid") UUID nodeUuid,
             @Parameter(description = "Should get in upstream built node ?") @RequestParam(value = "inUpstreamBuiltParentNode", required = false, defaultValue = "false") boolean inUpstreamBuiltParentNode,
             @RequestBody String filter) {
-
         return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studyService.evaluateFilter(studyUuid, nodeUuid, inUpstreamBuiltParentNode, filter));
     }
+
+    @GetMapping(value = "/studies/{studyUuid}/filter/export")
```
FranckLecuyer, line 1933: "/studies/{studyUuid}/filters/{filterUuid}/export "  ?

src/main/java/org/gridsuite/study/server/service/FilterService.java
```diff
@@ -77,4 +78,15 @@ public String evaluateFilter(UUID networkUuid, String variantId, String filter)
             }
         }
     }
+
+    public String exportFilter(UUID networkUuid, UUID filterUuid) {
+        Objects.requireNonNull(networkUuid);
```
FranckLecuyer, line 83: Objects.requireNonNull(filterUuid);   also

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/study/server/StudyController.java
dbraquart: Yes, I was hesitating because a filter is not directly a sub-resource of a study.
But you're right, it's better to use the Path.
And I want to replace export by elements, this is easier to understand.

Meklo (approved):

FranckLecuyer (approved):

PR #538, opened 2024-03-28
feat(voltage-init-parameters): add parameters related to study and voltage-init

TheMaskedTurtle:


src/main/java/org/gridsuite/study/server/repository/StudyEntity.java
```diff
@@ -172,6 +173,10 @@ public class StudyEntity extends AbstractManuallyAssignedIdentifierEntity<UUID>
     @Builder.Default
     private StudyIndexationStatus indexationStatus = StudyIndexationStatus.NOT_INDEXED;
 
+    @OneToOne(cascade = CascadeType.ALL, orphanRemoval = true)
+    @JoinColumn(name = "study_voltage_init_parameters_id")
```
FranckLecuyer, line 177: Set explicit and readable name for foreign key constraint

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/study/server/repository/StudyEntity.java
TheMaskedTurtle: I thought that was quite explicit already, it's parameters related to both study and voltage init parameters. Do you have any other idea ?

FranckLecuyer: It's for the foreign key name
See securityAnalysisParameters in the same entity just above with JPA annotation : 
foreignKey = @ForeignKey(
                    name = "securityAnalysisParameters_id_fk"
            ))

FranckLecuyer (changes requested):

src/main/java/org/gridsuite/study/server/repository/StudyEntity.java
TheMaskedTurtle: ok fixed !

FranckLecuyer (approved):

PR #546, opened 2024-04-12
feat: add error message on node build failed

AbdelHedhili:


src/main/java/org/gridsuite/study/server/notification/NotificationService.java
```diff
@@ -241,6 +241,16 @@ public void emitNodesChanged(UUID studyUuid, Collection<UUID> nodes) {
         );
     }
 
+    @PostCompletion
+    public void emitNodeBuildFailed(UUID studyUuid, UUID nodeUuid, String errorMessage) {
+        sendUpdateMessage(MessageBuilder.withPayload("")
+                .setHeader(HEADER_STUDY_UUID, studyUuid)
+                .setHeader(HEADER_NODE, nodeUuid)
+                .setHeader(HEADER_UPDATE_TYPE, UPDATE_TYPE_BUILD_FAILED)
+                .setHeader(HEADER_ERROR, errorMessage)
```
dbraquart, line 250: it's a pitie not to be able to emit the userId too. But it looks like the modification-server does not have/sent it.

dbraquart (approved):
code: ok
tests: ok
remark: the errorMessage is not limited to the user that has ran the built.

PR #550, opened 2024-04-18
Refactor elements endpoint

jamal-khey:
This pull request refactors the elements endpoint to include additional parameters in the URL search params. Previously, the parameters that are 'specific' to an equipement type are mixed with generic params for all equipements


Mathieu-Deharbe (changes requested):
The "operation" parameter seems to be lost when I create a shunt compensator network modification and open it. The following request is called : /network/elements/.ENGS%207%20COND.1?inUpstreamBuiltParentNode=true&elementType=SHUNT_COMPENSATOR&infoType=FORM&additionalParams%5BMODIFICATION%5D=operation

Which is fine, but then the "operation" parameter is lost in this endpoint of study-server: 
```
@GetMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/network/elements/{elementId}")
    @Operation(summary = "Get network elements infos")
    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "The list of network elements infos")})
    public ResponseEntity<String> getNetworkElementInfos(
            @Parameter(description = "Study uuid") @PathVariable("studyUuid") UUID studyUuid,
            @Parameter(description = "Node uuid") @PathVariable("nodeUuid") UUID nodeUuid,
            @Parameter(description = "Element id") @PathVariable("elementId") String elementId,
            @Parameter(description = "Element type") @RequestParam(name = "elementType") String elementType,
            @Parameter(description = "Info type") @RequestParam(name = "infoType") String infoType,
            @Parameter(description = "Operation") @RequestParam(name = "operation", required = false) String operation,
            @Parameter(description = "Should get in upstream built node ?") @RequestParam(value = "inUpstreamBuiltParentNode", required = false, defaultValue = "false") boolean inUpstreamBuiltParentNode) {

        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studyService.getNetworkElementInfos(studyUuid, nodeUuid, elementType, infoType, elementId, operation, inUpstreamBuiltParentNode));
    }
```

jamal-khey:
> /studies/{studyUuid}/nodes/{nodeUuid}/network/elements/
@Mathieu-Deharbe i fixed the endpoint , now the params are not lost

src/main/java/org/gridsuite/study/server/dto/InfoTypeParameters.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+@Data
+@Builder
+public class InfoTypeParameters {
+    public static final String QUERY_PARAM_DC_POWERFACTOR = "dcPowerFactor";
+    public static final String QUERY_PARAM_OPERATION = "operation";
+    private String infoType;
+    private Map<String, String> optionalParameters;
```
Mathieu-Deharbe, line 21: I find that this way of putting together these two data -which are not even the only parameters- quite artificial. The resulting `InfoTypeParameters` is a class name that looks more like a collage than a real data structure.

Mathieu-Deharbe (approved):

PR #559, opened 2024-05-03
Add short circuit parameter

souissimai:


src/main/java/org/gridsuite/study/server/dto/ShortCircuitPredefinedConfiguration.java
```diff
@@ -11,5 +11,6 @@
  */
 public enum ShortCircuitPredefinedConfiguration {
     ICC_MAX_WITH_NOMINAL_VOLTAGE_MAP,
-    ICC_MAX_WITH_CEI909
+    ICC_MAX_WITH_CEI909,
+    ICC_MIN_WITH_NOMINAL
```
AAJELLAL, line 15: I think you should change this. **ICC_MIN_WITH_NOMINAL_VOLTAGE_MAP** is more  convenient.

AAJELLAL (changes requested):
A name change + You need to add one more test with the predefinedParameters  **ICC_MIN_WITH_NOMINAL**

PR #563, opened 2024-05-14
bugfix, don't reset provider for loadflow param reset from profile

jonenst:
This is for consistency with how it works from system default values


ayolab (approved):

dbraquart (approved):
code: ok
tests: ok

PR #564, opened 2024-05-14
Feature: make getDefaultLoadFlowProvider return provider from profile when available

jonenst:


ayolab (approved):

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -802,7 +802,19 @@ public boolean setLoadFlowParameters(UUID studyUuid, String parameters, String u
...
+    public String getDefaultLoadflowProvider(String userId) {
+        if (userId != null) {
+            UserProfileInfos userProfileInfos = userAdminService.getUserProfile(userId).orElse(null);
+            if (userProfileInfos != null && userProfileInfos.getLoadFlowParameterId() != null) {
+                try {
+                    return loadflowService.getLoadFlowParameters(userProfileInfos.getLoadFlowParameterId()).getProvider();
+                } catch (Exception e) {
+                    LOGGER.error(String.format("Could not get loadflow parameters with id '%s' from user/profile '%s/%s'. Using default parameters",
+                            userProfileInfos.getLoadFlowParameterId(), userId, userProfileInfos.getName()), e);
+                    // in case of duplication error (ex: wrong/dangling uuid in the profile), move on with default params below
```
dbraquart, line 814: remove comment

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -802,7 +802,19 @@ public boolean setLoadFlowParameters(UUID studyUuid, String parameters, String u
...
 
-    public String getDefaultLoadflowProvider() {
+    public String getDefaultLoadflowProvider(String userId) {
+        if (userId != null) {
+            UserProfileInfos userProfileInfos = userAdminService.getUserProfile(userId).orElse(null);
+            if (userProfileInfos != null && userProfileInfos.getLoadFlowParameterId() != null) {
+                try {
+                    return loadflowService.getLoadFlowParameters(userProfileInfos.getLoadFlowParameterId()).getProvider();
+                } catch (Exception e) {
+                    LOGGER.error(String.format("Could not get loadflow parameters with id '%s' from user/profile '%s/%s'. Using default parameters",
```
dbraquart, line 812:
```suggestion
                    LOGGER.error(String.format("Could not get loadflow parameters with id '%s' from user/profile '%s/%s'. Using default provider",
```

dbraquart (changes requested):
code: ok, just a comment to remove
tests: ko, not conform
- "Reset All" button : works as expected
- "Reset provider default values":
  - with no profile : ok
  - with a profile: ko. we still restore all params stored in the profile, while we should set the provider dft values (to be confirmed with the PO ?)

dbraquart (approved):
code: ok
tests: ok

PR #576, opened 2024-05-27
re-indent file

Tristan-WorkGH:
Just reindent code, no code modification.

deported refactor from #569 to lighten the diff


Tristan-WorkGH:
Won't be merged as no rules define for that currently.

PR #578, opened 2024-05-31
Add the possibility to launch a state estimation on a network

FranckLecuyer:


src/test/java/org/gridsuite/study/server/NetworkModificationTest.java
```diff
@@ -126,6 +126,9 @@ public class NetworkModificationTest {
 
     private static final String VOLTAGE_INIT_STATUS_JSON = "{\"status\":\"COMPLETED\"}";
 
+    private static final String STATE_ESTIMATION_RESULT_UUID = "d3a85e9b-9894-4255-8ec7-07ea965d24eb";
+    private static final String STATE_ESTIMATION_STATUS_JSON = "\"COMPLETED\"";
```
anistouri, line 130: Why is it not consistent with the other calculations?
{\"status\":\"COMPLETED\"}

dbraquart: "COMPLETED" is what is returned by the estim computation server !
I guess we should change all the others to be consistant ;-)

dbraquart (approved):
code: ok
tests: ok

anistouri (approved):
Code OK

PR #581, opened 2024-06-10
Fix wrong content-type for SVG

Tristan-WorkGH:


src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -931,7 +931,7 @@ public ResponseEntity<ShortCircuitParametersInfos> getShortCircuitParameters(
         return ResponseEntity.ok().body(studyService.getShortCircuitParametersInfo(studyUuid));
     }
 
-    @GetMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/network/substations/{substationId}/svg")
+    @GetMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/network/substations/{substationId}/svg", produces = IMAGE_SVG_VALUE)
```
sBouzols, line 934: Add `produces = IMAGE_SVG_VALUE` as well on `getVoltageLevelDiagram` ?

PR #583, opened 2024-06-10
fix: search for equipment recreated after deletion

anistouri:


src/test/java/org/gridsuite/study/server/StudyTest.java
```diff
@@ -737,6 +738,19 @@ public void testSearch() throws Exception {
                 .andReturn();
     }
 
+    @Test
+    public void cleanRemovedEquipmentsInSerachTest() {
```
achour94, line 742: This test should be done in EquipmentInfosServiceTests

achour94 (approved):

PR #595, opened 2024-07-15
add a profile quota to limit node builds number

dbraquart:


achour94 (approved):

src/main/java/org/gridsuite/study/server/service/NetworkModificationTreeService.java
```diff
@@ -1011,4 +1012,17 @@ private Optional<UUID> doGetParentNode(UUID nodeUuid, NodeType nodeType) {
             return doGetParentNode(nodeEntity.getParentNode().getIdNode(), nodeType);
         }
     }
+
+    public int countBuiltNodes(UUID studyUuid) {
+        // retrieve all network modifications un-stashed nodes existing in this study
+        List<NodeEntity> nodes = nodesRepository.findAllByStudyIdAndTypeAndStashed(studyUuid, NodeType.NETWORK_MODIFICATION, false);
+        AtomicInteger builtNodeNumber = new AtomicInteger(0);
```
achour94, line 1019: Why use AtomicInteger here? It seems unnecessary since this method is not called in a multi-threaded context.

dbraquart: you cannot modify a local variable inside a lambda (I could have used a for loop instead)

achour94: `
(int) nodes.stream()
                .filter(n -> repositories.get(n.getType()).getNodeBuildStatus(n.getIdNode()).isBuilt())
                .count();
`

achour94: you can try this ?

dbraquart: yes it works, and it's prettier

PR #599, opened 2024-08-01
adds optional filename parameter for case export

Mathieu-Deharbe:


src/main/java/org/gridsuite/study/server/service/NetworkConversionService.java
```diff
@@ -92,15 +92,23 @@ public String getExportFormats() {
...
         if (!variantId.isEmpty()) {
             uriComponentsBuilder.queryParam("variantId", variantId);
         }
 
-        if (!StringUtils.isEmpty(studyName) && !StringUtils.isEmpty(nodeName)) {
-            uriComponentsBuilder.queryParam("fileName", URLEncoder.encode(studyName + "_" + nodeName, StandardCharsets.UTF_8));
+        String finalFilename = fileName;
+        if (StringUtils.isEmpty(finalFilename) && !StringUtils.isEmpty(studyName) && !StringUtils.isEmpty(nodeName)) {
+            finalFilename = URLEncoder.encode(studyName + "_" + nodeName, StandardCharsets.UTF_8);
+        }
```
thangqp, line 109: In my opinion, studyName and nodeName now have no role, a fileName is enough. Too many optional params for the same objective make the code difficult to maintain.

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -780,9 +780,10 @@ public ResponseEntity<byte[]> exportNetwork(
             @PathVariable("nodeUuid") UUID nodeUuid,
             @PathVariable("format") String format,
             @RequestParam(value = "formatParameters", required = false) String parametersJson,
-            @RequestParam(value = "studyName", required = false) String studyName) {
+            @RequestParam(value = "studyName", required = false) String studyName,
```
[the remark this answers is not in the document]
souissimai: If the user does not provide a filename, the file should be automatically named NameNode_NameStudyInDirectory. To handle this, the front-end should check if the filename parameter is provided. If it is, there's no need to pass the studyName. If not, the network conversion server should construct the default filename.

src/main/java/org/gridsuite/study/server/service/NetworkConversionService.java
souissimai: I think we need to construct it in backed for case of zip file exported, to be tested

souissimai (approved):

src/main/java/org/gridsuite/study/server/StudyController.java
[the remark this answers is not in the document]
Mathieu-Deharbe: But if this endpoint is only used in this case : https://github.com/gridsuite/gridstudy-app/pull/2196 fileName will always be used (because the default .value NameNode_NameStudyInDirectory is generated and turned into fileName on the front side)

Actually I may make the _filename_ parameter required.

src/main/java/org/gridsuite/study/server/service/NetworkConversionService.java
Mathieu-Deharbe: I didn't find how to export a zip file (and use this function).

thangqp (approved):
Code OK, Test OK

ghazwarhili (approved):

PR #600, opened 2024-08-05
add limit reductions

souissimai:


src/main/java/org/gridsuite/study/server/dto/LimitReductionsByVoltageLevel.java
```diff
@@ -0,0 +1,51 @@
+/**
```
SlimaneAmar, line 1: Remove DTO
Pass a string instead !

SlimaneAmar (changes requested):

SlimaneAmar (approved):

PR #602, opened 2024-08-13
Update Report dto

ayolab:


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1639,7 +1639,7 @@ public Report getSubReport(String subReportId, Set<String> severityLevels) {
 
     @Transactional(readOnly = true)
     public List<Report> getNodeReport(UUID nodeUuid, String reportId, ReportType reportType, Set<String> severityLevels) {
-        return reportService.getReport(UUID.fromString(reportId), nodeUuid.toString(), nodeUuid + "@" + reportType.reportKey, ReportNameMatchingType.EXACT_MATCHING, severityLevels);
+        return reportService.getReport(UUID.fromString(reportId), nodeUuid.toString(), null, null, severityLevels);
```
TheMaskedTurtle, line 1642: Why do you change this behavior ?

ayolab: no need for filtering here we already call this endpoint with the right report ID we want to access (the one related to NETWORK_MODIFICATIONS / COMPUTATIONS)

TheMaskedTurtle (approved):

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1639,7 +1639,7 @@ public Report getSubReport(String subReportId, Set<String> severityLevels) {
 
     @Transactional(readOnly = true)
     public List<Report> getNodeReport(UUID nodeUuid, String reportId, ReportType reportType, Set<String> severityLevels) {
-        return reportService.getReport(UUID.fromString(reportId), nodeUuid.toString(), nodeUuid + "@" + reportType.reportKey, ReportNameMatchingType.EXACT_MATCHING, severityLevels);
```
dbraquart, line 1642: reportType is not used anymore: can be remove from here, rest api, and front-end

dbraquart (changes requested):
code: some clean can be done
tests: on going

dbraquart (approved):
code: ok
tests: ok

PR #603, opened 2024-08-21
add static var compensator modification type

ghazwarhili:


thangqp:
I don't see the role of class ModificationInfos except it is used in test classes, but you do not add any test for static var compensator

PR #613, opened 2024-09-11
upgrade network store server to v1.16.0

EtienneLt:


pom.xml
```diff
@@ -132,6 +133,7 @@
         <dependency>
             <groupId>com.powsybl</groupId>
             <artifactId>powsybl-network-store-client</artifactId>
+            <version>${powsybl-network-store.version}</version>
```
SlimaneAmar, line 136: Need `powsybl-network-store-iidm-impl` and `powsybl-network-store-model` ?

FranckLecuyer (approved):

PR #614, opened 2024-09-11
Use notifications to manage parameter changes

AAJELLAL:

PR #620, opened 2024-10-08
Extract computations notification types

Mathieu-Deharbe:
removes a Circular dependency from https://github.com/gridsuite/study-server/pull/614


SlimaneAmar:
I think it's an anti pattern to in a constants class
[https://www.baeldung.com/java-constants-good-practices](https://www.baeldung.com/java-constants-good-practices)

Mathieu-Deharbe:
I don't think so : the anti-pattern described in your article is "A Large Global Constants Class" which contains everything. Here I have a "type" (ComputationsNotificationTypes) for the constant class. The NotificationService with its dozens of constants was much closer to the anti pattern in my opinion.

Look a the good pattern he describes at the end : "4.2. Constants Class"

If you prefer I can put them into ComputationType. But given that this is an enum and not a class I would have to add a nested class like this : 
https://stackoverflow.com/questions/23608885/how-to-define-static-constants-in-a-java-enum

PR #624, opened 2024-10-15
[WIP] use spring boot autoconfigure instead of our hand rolled esconfig

jonenst:
this allows to use multiple addresses


jonenst:
TODO: check if we can get the hosts from a different place than boot.autoconfigure.ElasticsearchConnectionDetails, so that if we stop using autoconfigure we still get the information. Does the final client expose this ?

SlimaneAmar:
Replaced by https://github.com/gridsuite/study-server/pull/646

PR #649, opened 2024-12-12
Use study-config-server to store network visualization parameters per study

dbraquart:


src/main/java/org/gridsuite/study/server/service/StudyConfigService.java
```diff
@@ -0,0 +1,141 @@
+/**
+ * Copyright (c) 2021, RTE (http://www.rte-france.com)
```
ayolab, line 2:
```suggestion
 * Copyright (c) 2024, RTE (http://www.rte-france.com)
```

ayolab (changes requested):

ayolab (approved):

PR #650, opened 2024-12-12
Migrate security analysis endpoints to add rootNetworkUuid in path

klesaulnier:


src/test/java/org/gridsuite/study/server/utils/StudyTestUtils.java
```diff
@@ -0,0 +1,21 @@
+package org.gridsuite.study.server.utils;
```
SlimaneAmar, line 1: Why we not use `TestUtils` ?

SlimaneAmar:
Throw an exception for this method is ok    
```
public UUID getNetworkUuid(UUID rootNetworkUuid) {
        return getRootNetwork(rootNetworkUuid).map(RootNetworkEntity::getNetworkUuid).orElseThrow(() -> new StudyException(NETWORK_NOT_FOUND));
    }
```

SlimaneAmar (approved):

PR #662, opened 2025-01-03
Retrieve all tombstoned in the list instead of a sublist to deduce the equipments that are present

antoinebhs:
In the current implementation, we retrieve the first 400 tombstoned equipments to filter the equipments returned by ES. This is completely wrong if there are more than 400 tombstoned equipments. 
I changed this to retrieve the tombstoned equipments among the equipments returned by ES.


src/main/java/org/gridsuite/study/server/elasticsearch/EquipmentInfosService.java
```diff
@@ -120,6 +120,13 @@ public List<TombstonedEquipmentInfos> findAllTombstonedEquipmentInfos(@NonNull U
         return tombstonedEquipmentInfosRepository.findAllByNetworkUuid(networkUuid);
     }
 
+    public Set<TombstonedEquipmentInfos> findTombstonedEquipmentInfosByIdIn(@NonNull UUID networkUuid, @NonNull String variantId, @NonNull List<String> equipmentIds) {
+        return tombstonedEquipmentInfosRepository.findByIdInAndNetworkUuidAndVariantId(
```
SlimaneAmar, line 124: `findAllByIdAndNetworkUuidAndVariantId` ?

src/main/java/org/gridsuite/study/server/elasticsearch/EquipmentInfosService.java
```diff
@@ -120,6 +120,13 @@ public List<TombstonedEquipmentInfos> findAllTombstonedEquipmentInfos(@NonNull U
         return tombstonedEquipmentInfosRepository.findAllByNetworkUuid(networkUuid);
     }
 
+    public Set<TombstonedEquipmentInfos> findTombstonedEquipmentInfosByIdIn(@NonNull UUID networkUuid, @NonNull String variantId, @NonNull List<String> equipmentIds) {
```
SlimaneAmar, line 123: private ?

src/main/java/org/gridsuite/study/server/elasticsearch/EquipmentInfosService.java
```diff
@@ -120,6 +120,13 @@ public List<TombstonedEquipmentInfos> findAllTombstonedEquipmentInfos(@NonNull U
...
+        return tombstonedEquipmentInfosRepository.findByIdInAndNetworkUuidAndVariantId(
+                        equipmentIds,
+                        networkUuid,
+                        variantId);
+    }
+
```
SlimaneAmar, line 129: Perhaps some performance issues with big lists ?

src/main/java/org/gridsuite/study/server/elasticsearch/EquipmentInfosService.java
antoinebhs: If you do that, you can only pass one equipmentId, I think we want to pass a list here

src/main/java/org/gridsuite/study/server/elasticsearch/EquipmentInfosService.java
antoinebhs: Fixed in https://github.com/gridsuite/study-server/pull/662/commits/b275ff84a755a1a78e6907051a6188b0a116e481

SlimaneAmar (approved):

src/main/java/org/gridsuite/study/server/elasticsearch/EquipmentInfosService.java
antoinebhs: Actually, it's needed for tests. I reverted to public in https://github.com/gridsuite/study-server/pull/662/commits/4a4f54ce233dfc0b2469ea4b1b67f105356f5777

antoinebhs: EquipmentInfosService is a MockedBean in some tests (in StudyTest.java). findTombstonedEquipmentInfosByIdIn calls the tombstonedEquipmentInfosRepository and need to be mocked in the tests. If it's private, it's not mocked which leads to an error, that's why it needs to be public as all methods in this service calling repositories.

PR #669, opened 2025-01-30
add subtype

ghazwarhili:


src/main/java/org/gridsuite/study/server/dto/elasticsearch/EquipmentInfos.java
```diff
@@ -49,6 +49,9 @@ public class EquipmentInfos extends BasicEquipmentInfos {
     @Field("equipmentType")
     String type;
 
+    @Field(type = FieldType.Keyword)
+    Set<String> equipmentSubType;
```
thangqp, line 53:
```suggestion
    Set<String> equipmentSubTypes;
```
Since it is a collection, should add 's' to the field name if the name has a same semantics

src/main/java/org/gridsuite/study/server/elasticsearch/EquipmentInfosService.java
```diff
@@ -79,6 +79,7 @@ public enum FieldSelector {
     static final String EQUIPMENT_NAME = "equipmentName.fullascii";
     static final String EQUIPMENT_ID = "equipmentId.fullascii";
     static final String EQUIPMENT_TYPE = "equipmentType.keyword";
+    static final String EQUIPMENT_SUB_TYPE = "equipmentSubType.keyword";
```
thangqp, line 82:
```suggestion
    static final String EQUIPMENT_SUB_TYPE = "equipmentSubTypes.keyword";
```

PR #672, opened 2025-01-31
Fix filter conversion - when selecting a study, first root network is now selected

klesaulnier:


SlimaneAmar (approved):

src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -2068,6 +2068,15 @@ public ResponseEntity<String> exportFilter(
...
+    @GetMapping(value = "/studies/{studyUuid}/filters/{filterUuid}/elements")
+    @Operation(summary = "Evaluate a filter on root node to get matched elements")
+    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "The list of matched elements")})
+    public ResponseEntity<String> exportFilterFromFirstRootNetwork(
+        @Parameter(description = "Study uuid") @PathVariable("studyUuid") UUID studyUuid,
+        @Parameter(description = "Filter uuid to be applied") @PathVariable("filterUuid") UUID filterUuid) {
+        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studyService.exportFilter(studyService.getStudyFirstRootNetworkUuid(studyUuid), filterUuid));
+    }
+
     @GetMapping(value = "/studies/{studyUuid}/root-networks/{rootNetworkUuid}/filters/elements")
```
souissimai, line 2080: We were supposed to use the @GetMapping(value = "/studies/{studyUuid}/root-networks/{rootNetworkUuid}/filters/elements") endpoint, but since the frontend cannot select and pass a rootNetwork, a new endpoint was created to pass the firstRootNetwork instead: this endpoint is not used !

PR #676, opened 2025-02-10
Fix filter conversion - when selecting a study, first root network is now selected

klesaulnier:


EstherDarkish (approved):
Ok since already approved in https://github.com/gridsuite/study-server/pull/672/

PR #682, opened 2025-02-13
Fix supervision endpoints error with root networks in creation

antoinebhs:
When there are root networks in creation in the table **root_network_creation_request** (we need to investigate why there are some rows that remain here and never get deleted => crash of pod, error on import, etc), we try to invalidate node/delete index on these networks that don't exist yet which throw an error "ROOT_NETWORK_NOT_FOUND" as it does not exist.

In this PR, I propose to use only existing root network for supervision endpoints: to invalidate nodes and delete indexed equipments.

### DIFF
--- a/src/main/java/org/gridsuite/study/server/service/StudyService.java
+++ b/src/main/java/org/gridsuite/study/server/service/StudyService.java
@@ -2503,17 +2503,21 @@ public class StudyService {
         return rootNetworkEntities.stream().map(RootNetworkEntity::toDto).toList();
     }
 
-    @Transactional
+    @Transactional(readOnly = true)
     public List<BasicRootNetworkInfos> getAllBasicRootNetworkInfos(UUID studyUuid) {
         return Stream
             .concat(
-                getStudyRootNetworks(studyUuid).stream().map(RootNetworkEntity::toBasicDto),
+                getExistingRootNetworkInfos(studyUuid).stream(),
                 rootNetworkService.getCreationRequests(studyUuid).stream().map(RootNetworkCreationRequestEntity::toBasicDto))
             .toList();
     }
 
-    @Transactional
+    @Transactional(readOnly = true)
     public List<BasicRootNetworkInfos> getExistingBasicRootNetworkInfos(UUID studyUuid) {
+        return getExistingRootNetworkInfos(studyUuid);
+    }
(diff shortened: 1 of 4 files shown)


klesaulnier (approved):

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2504,14 +2504,19 @@ private List<RootNetworkInfos> getStudyRootNetworksInfos(UUID studyUuid) {
     }
 
     @Transactional
-    public List<BasicRootNetworkInfos> getBasicRootNetworkInfos(UUID studyUuid) {
+    public List<BasicRootNetworkInfos> getAllBasicRootNetworkInfos(UUID studyUuid) {
```
SlimaneAmar, line 2507: @Transactional(readOnly = true) ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2504,14 +2504,19 @@ private List<RootNetworkInfos> getStudyRootNetworksInfos(UUID studyUuid) {
...
         return Stream
             .concat(
                 getStudyRootNetworks(studyUuid).stream().map(RootNetworkEntity::toBasicDto),
                 rootNetworkService.getCreationRequests(studyUuid).stream().map(RootNetworkCreationRequestEntity::toBasicDto))
             .toList();
     }
 
+    @Transactional
```
SlimaneAmar, line 2515: @Transactional(readOnly = true) ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
SlimaneAmar, line 2507: Call `getExistingBasicRootNetworkInfos` ?

SlimaneAmar (approved):

src/main/java/org/gridsuite/study/server/service/StudyService.java
antoinebhs: Fixed in https://github.com/gridsuite/study-server/pull/682/commits/eac5ac5b6f6e63a69abe2f5ca46894a4eaddc8a7

PR #689, opened 2025-02-21
Add tag for root network

sBouzols:


souissimai (commented):
For  already created RRs, shouldn't there be a default tag in the database? I think we need to migrate?

AAJELLAL (approved):
Tests and code are OK.

souissimai (approved):
Code ok, TESTTED

PR #711, opened 2025-03-12
WIP New API entry to save a Network Area Diagram config

EstherDarkish:


src/main/java/org/gridsuite/study/server/StudyController.java
```diff
@@ -1116,6 +1118,19 @@ public ResponseEntity<String> getNeworkAreaDiagram(
             ResponseEntity.noContent().build();
     }
 
+    @PostMapping(value = "/studies/{studyUuid}/root-networks/{rootNetworkUuid}/nodes/{nodeUuid}/network-area-diagram/config", consumes = MediaType.APPLICATION_JSON_VALUE)
```
antoinebhs, line 1121: Not sure why we need this? Can't we call directly single line diagram server from explore server?

antoinebhs (commented):
First comment reading the code

src/main/java/org/gridsuite/study/server/StudyController.java
EstherDarkish: Single-line-diagram-server needs the variantId and the only way to get it is from study-server. We do not want this variantId to be visible in the frontend.
This solution is the one we choosed by collectively pondering our options (and with a TL).

klesaulnier (changes requested):
Remove DTOs from study-server and pass body as string only since study-server does not need to know them, it only transfer them to the next service

The rest of the code seems to be ok

EstherDarkish:
Closing this PR as the design changed. This is no longer needed.

PR #727, opened 2025-03-31
Remove node associated to an alias on node stash

FranckLecuyer:


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1840,6 +1839,12 @@ public void deleteNodes(UUID studyUuid, List<UUID> nodeIds, boolean deleteChildr
                 }
             });
         }
+    }
+
+    @Transactional
+    public void deleteNodes(UUID studyUuid, List<UUID> nodeIds, boolean deleteChildren, String userId) {
+        removeNodesFromAliases(studyUuid, nodeIds);
```
dbraquart, line 1846: I cant see why we need this, because alias update is already made when we stash a node.

dbraquart (commented):
code: juste a question about deleteNodes
tests: ok

src/main/java/org/gridsuite/study/server/service/StudyService.java
FranckLecuyer: If we just call manually the endpoint that delete nodes (with a curl ...) without having called the endpoint that stash nodes ...

dbraquart (approved):
tests OK

dbraquart (changes requested):
test ko on single node removal

dbraquart (approved):

PR #738, opened 2025-04-28
Pass from monoroot to multiroot study

souissimai:


src/main/java/org/gridsuite/study/server/repository/StudyEntity.java
```diff
@@ -129,6 +129,9 @@ public class StudyEntity extends AbstractManuallyAssignedIdentifierEntity<UUID>
         ))
     private List<NodeAliasEmbeddable> nodeAliases;
 
+    @Column(name = "mono_root", columnDefinition = "boolean default true")
+    private boolean monoRoot;
```
klesaulnier, line 133:
```suggestion
    @Column(name = "mono_root")
    private boolean monoRoot = true;
```

then deal with default value in database with liquibase changelog

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -339,6 +341,10 @@ public void createRootNetwork(@NonNull UUID studyUuid, @NonNull RootNetworkInfos
             rootNetworkInfos.setTag(rootNetworkCreationRequestEntityOpt.get().getTag());
             rootNetworkService.createRootNetwork(studyEntity, rootNetworkInfos);
             rootNetworkService.deleteRootNetworkRequest(rootNetworkCreationRequestEntityOpt.get());
+            //update study entity to multi root
+            if (studyEntity.isMonoRoot() && studyEntity.getRootNetworks().size() > 1) {
```
klesaulnier, line 345: do we need to check this ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1321,6 +1327,7 @@ private StudyEntity saveStudyThenCreateBasicTree(UUID studyUuid, NetworkInfos ne
                 .dynamicSecurityAnalysisParametersUuid(dynamicSecurityAnalysisParametersUuid)
                 .stateEstimationParametersUuid(stateEstimationParametersUuid)
                 .spreadsheetConfigCollectionUuid(spreadsheetConfigCollectionUuid)
+                .monoRoot(true)
```
klesaulnier, line 1330: unecessary if you add @Builder.default on "monoRoot" field

src/main/java/org/gridsuite/study/server/service/StudyService.java
souissimai: We should update the flag to false only when transitioning from a single root network (mono root) to multiple root networks when adding the secon root network

src/main/java/org/gridsuite/study/server/repository/StudyEntity.java
souissimai: The default value depends on the number of root networks in the study. Based on this, we can determine the flag’s state (true or false) that I did with update changeset , no need to have default value I think *

souissimai: already discussed

src/main/java/org/gridsuite/study/server/service/StudyService.java
souissimai: already discussed

src/main/java/org/gridsuite/study/server/service/StudyService.java
souissimai: default value is not true as discussed

AAJELLAL (approved):
Test and code are ok.

PR #741, opened 2025-04-30
network export copy content-length when available to restore downloads progressbar estimates

jonenst:


antoinebhs (approved):
Code OK
cc @sBouzols

sBouzols (approved):

PR #747, opened 2025-05-15
Make root networks limits configurable in application.yml

jonenst:
Stopgap until we put it in the profile or the study configuration


klesaulnier (commented):
root_network tag being a varchar(4), I don't think we should externalize "max-root-network-tag-length"

klesaulnier (approved):

jonenst:
front still prevents going to more than 3 but we can work on that if needed

PR #749, opened 2025-05-19
On loadflow invalidation, invalidate node and all children

klesaulnier:


antoinebhs (approved):
Code OK
We can optimize invalidation on network modification creation to avoid applying then invalidate but only invalidate + create

PR #769, opened 2025-06-08
feat: add endpoint to update column states in spreadsheet configuration

achour94:


Tristan-WorkGH (changes requested):
You don't need that, the front can directly query the API of study-config-server.

Tristan-WorkGH (approved):
Okay, just 1 remark.

PR #770, opened 2025-06-10
fix filter export by adding node UUID and variant ID

achour94:


FranckLecuyer (approved):
Code review: OK
Tests: OK

PR #771, opened 2025-06-10
use simple

jonenst:
TODO do we have to do this ? Currently the apache httpcomponents has a limit of 5 connections per host which is top low for us, but when raising it I got NPEs


jonenst:
The NPEs were just from https://issues.apache.org/jira/browse/HTTPCLIENT-2313 because we had forced 5.3

jonenst:
superseeded by #774 which makes the study-server more homogenous with our other servers. This is not ideal because we have httpcomponents on the test classpath so during tests we don't have the same resttemplate underlying implementation as in production, but this should be worked on more globally on all our servers

PR #779, opened 2025-06-17
Invalidate nodes when invalidate all loadflow from supervision

klesaulnier:


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1301,7 +1301,7 @@ private void invalidateLoadFlowStatusOnAllNodes(UUID studyUuid) {
         loadflowService.invalidateLoadFlowStatus(rootNetworkNodeInfoService.getComputationResultUuids(studyUuid, LOAD_FLOW));
     }
 
-    private void invalidateNodeTreeWithLoadFlowResults(UUID studyUuid) {
+    public void invalidateNodeTreeWithLoadFlowResults(UUID studyUuid) {
```
SlimaneAmar, line 1304: This punlic method is called by a private method !

klesaulnier: I agree, but what it the best workaround here

src/main/java/org/gridsuite/study/server/service/SupervisionService.java
```diff
@@ -163,14 +165,8 @@ private Integer deleteLoadflowResults() {
...
         List<RootNetworkNodeInfoEntity> rootNetworkNodeInfoEntities = rootNetworkNodeInfoRepository.findAllByLoadFlowResultUuidNotNull();
-        List<UUID> reportsToDelete = new ArrayList<>();
-        rootNetworkNodeInfoEntities.forEach(rootNetworkNodeInfo -> {
-            rootNetworkNodeInfo.setLoadFlowResultUuid(null);
-            reportsToDelete.add(rootNetworkNodeInfo.getComputationReports().get(ComputationType.LOAD_FLOW.name()));
-            rootNetworkNodeInfo.getComputationReports().remove(ComputationType.LOAD_FLOW.name());
-        });
-        reportService.deleteReports(reportsToDelete);
-        loadFlowService.deleteAllLoadFlowResults();
+        List<UUID> studyUuids = rootNetworkNodeInfoEntities.stream().map(rnnie -> rnnie.getRootNetwork().getStudy().getId()).distinct().toList();
```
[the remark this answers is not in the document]
klesaulnier: I think it's to keep the previous optimization 
Otherwise, for each study in the app, we will load study by study rootNetworkNodeInfos having LF result, even if the study does not have any

src/main/java/org/gridsuite/study/server/service/StudyService.java
klesaulnier: fixed

SlimaneAmar (approved):

PR #812, opened 2025-07-30
Handle the nad generation mode

AAJELLAL:


src/main/java/org/gridsuite/study/server/service/SingleLineDiagramService.java
```diff
@@ -248,4 +249,18 @@ public void addParameters(DiagramParameters diagramParameters, UriComponentsBuil
             uriComponentsBuilder.queryParam(QUERY_PARAM_VARIANT_ID, variantId);
         }
     }
+
+    public UUID createDiagramConfig(String diagramConfig) {
```
EstherDarkish, line 253: There is a createDiagramConfig line 197. What about it ?

SlimaneAmar (approved):

PR #816, opened 2025-08-01
Feat: Manage Map Card save in layout

sBouzols:
feat(DiagramGridLayout): Manage Map Card save in layout and add a control over the number of Map Card in the grid


EstherDarkish (approved):
Code OK Test OK

PR #819, opened 2025-08-06
Simplify modification reports management

ayolab:


src/main/java/org/gridsuite/study/server/service/RootNetworkNodeInfoService.java
```diff
@@ -254,7 +252,6 @@ public InvalidateNodeInfos invalidateRootNetworkNode(RootNetworkNodeInfoEntity r
         InvalidateNodeInfos invalidateNodeInfos = getInvalidationComputationInfos(rootNetworkNodeInfoEntity, invalidateTreeParameters.computationsInvalidationMode());
 
         if (notOnlyChildrenBuildStatus) {
-            rootNetworkNodeInfoEntity.getModificationReports().forEach((key, value) -> invalidateNodeInfos.addReportUuid(value));
```
TheMaskedTurtle, line 257: That's kind of a regression, now we don't delete reports when we invalidate nodes. That's what we wanr in some cases but not for all cases. Maybe we should try to be more precise here. When no children are built we want to delete all the reports upward the next built node.

src/main/java/org/gridsuite/study/server/service/RootNetworkNodeInfoService.java
```diff
@@ -267,7 +264,6 @@ public InvalidateNodeInfos invalidateRootNetworkNode(RootNetworkNodeInfoEntity r
     private static void invalidateBuildStatus(RootNetworkNodeInfoEntity rootNetworkNodeInfoEntity, InvalidateNodeInfos invalidateNodeInfos) {
         rootNetworkNodeInfoEntity.setNodeBuildStatus(NodeBuildStatusEmbeddable.from(BuildStatus.NOT_BUILT));
         rootNetworkNodeInfoEntity.setVariantId(UUID.randomUUID().toString());
-        rootNetworkNodeInfoEntity.setModificationReports(new HashMap<>(Map.of(rootNetworkNodeInfoEntity.getNodeInfo().getId(), UUID.randomUUID())));
```
TheMaskedTurtle, line 270: Same case here

PR #828, opened 2025-08-27
save the nad associated to the user's profile with the newly created study

benrejebmoh:


dbraquart (changes requested):
Code: some refacto/renaming required
Tests: 1 failure
- In case of broken NAD link (in the profile), we create an invalid layout/diagram. We should just do nothing (leave the layout null). Should be fixed with the suggested modifications.

dbraquart (approved):
code: ok
tests: ok
Note: with a broken NAD link, we get an invalid NAD diagram in the study layout. But the PO stated this is acceptable.

PR #837, opened 2025-09-03
jorism/add-partial-loading

TheMaskedTurtle:


src/main/java/org/gridsuite/study/server/notification/NotificationService.java
```diff
@@ -201,6 +202,11 @@ public void emitSpreadsheetCollectionChanged(UUID studyUuid, UUID collectionUuid
         sendStudyUpdateMessage(studyUuid, UPDATE_SPREADSHEET_COLLECTION, MessageBuilder.withPayload(collectionUuid.toString()));
     }
 
+    @PostCompletion
+    public void emitSpreadsheetParametersChange(UUID studyUuid) {
+        sendStudyUpdateMessage(studyUuid, UPDATE_SPREADSHEET_PARAMETERS, MessageBuilder.withPayload(""));
```
Tristan-WorkGH, line 207: You didn't pass the new parameters to not have to request them from the front?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -858,9 +903,37 @@ public String getBranchOr3WTVoltageLevelId(UUID nodeUuid, UUID rootNetworkUuid,
...
+        additionalParameters.get(String.valueOf(ElementType.LINE)).put(
+            InfoTypeParameters.QUERY_PARAM_LOAD_OPERATIONAL_LIMIT_GROUPS,
+            String.valueOf(studyEntity.getSpreadsheetParameters().isSpreadsheetLoadLineOperationalLimitGroup()));
+        additionalParameters.get(String.valueOf(ElementType.TWO_WINDINGS_TRANSFORMER)).put(
+            InfoTypeParameters.QUERY_PARAM_LOAD_OPERATIONAL_LIMIT_GROUPS,
+            String.valueOf(studyEntity.getSpreadsheetParameters().isSpreadsheetLoadTwtOperationalLimitGroup()));
+        additionalParameters.put(String.valueOf(ElementType.GENERATOR),
+            Map.of(
+                InfoTypeParameters.QUERY_PARAM_LOAD_REGULATING_TERMINALS,
+                String.valueOf(studyEntity.getSpreadsheetParameters().isSpreadsheetLoadGeneratorRegulatingTerminal())));
```
Tristan-WorkGH, line 931: Isn't it over complicated to do a 2 level map instead of just 1 level ? the network-map-server can pick itself the parameter(s) it want.  
And we duplicate some parameter.

TheMaskedTurtle: It creates a bit more complexity here but that's the most flexible option for now, so that we don't have to hardcode every possible combinations of type/optionalParameter in the network-map-server. So I would still prefer this option.

dbraquart (changes requested):
code: just a typo to change
tests: works fine

dbraquart (approved):

PR #854, opened 2025-09-29
Remove code relating to the non evacuated energy computation

FranckLecuyer:


src/main/resources/db/changelog/changesets/changelog_20250925T151431Z.xml
```diff
@@ -0,0 +1,96 @@
+<?xml version="1.1" encoding="UTF-8" standalone="no"?>
+<databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog" xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext" xmlns:pro="http://www.liquibase.org/xml/ns/pro" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog-ext http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd http://www.liquibase.org/xml/ns/pro http://www.liquibase.org/xml/ns/pro/liquibase-pro-latest.xsd http://www.liquibase.org/xml/ns/dbchangelog http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+    <changeSet author="lecuyerfra (generated)" id="1758813303022-19">
+        <dropForeignKeyConstraint baseTableName="NON_EVACUATED_ENERGY_STAGES_SELECTION_DEFINITION_INDEX" constraintName="FK8FL4TQ030BR4BBKLVA1MM58N0"/>
```
TheMaskedTurtle, line 4: the constraint name is case sensitive and it does not work for me locally, I think you should rather remove these dropForeignKeyContraint and use cascadeConstraints="true" option in the the drop tables below

TheMaskedTurtle (changes requested):

TheMaskedTurtle (approved):

PR #978, opened 2026-04-15
Add invalidate study endpoint

Meklo:


src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
```diff
@@ -180,4 +179,12 @@ public ResponseEntity<Void> invalidateAllNodesBuilds(@PathVariable("studyUuid")
         return ResponseEntity.ok().build();
     }
 
+    @DeleteMapping(value = "/studies/{studyUuid}/unmount")
+    @Operation(summary = "Invalidate nodes builds and delete root node network")
+    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "study has been unmounted")})
+    public ResponseEntity<Void> unmountStudy(@PathVariable("studyUuid") UUID studyUuid) {
```
flomillot: Not sure if it's accurate, but in the ticket I see that

Image

src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
```diff
@@ -180,4 +179,12 @@ public ResponseEntity<Void> invalidateAllNodesBuilds(@PathVariable("studyUuid")
         return ResponseEntity.ok().build();
     }
 
+    @DeleteMapping(value = "/studies/{studyUuid}/unmount")
+    @Operation(summary = "Invalidate nodes builds and delete root node network")
```
flomillot:
```suggestion
    @Operation(summary = "Invalidate built nodes and delete root node network")
```
Also, what do you mean by 

> delete root node network

 ? When I look quickly to the code it's not clear to me what it does exactly.  Looks like it unbuild (invalidate?) nodes and delete all (or some?) root network data. 
https://github.com/gridsuite/study-server/blob/9e311186f283d39c6f4f7ea692f94020399e67e7/src/main/java/org/gridsuite/study/server/service/StudyService.java#L3941-L3943

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
```diff
@@ -235,18 +238,28 @@ public void deleteRootNetworks(StudyEntity studyEntity, Stream<UUID> rootNetwork
...
-    public void deleteRootNetworkRemoteInfos(List<RootNetworkInfos> rootNetworkInfos) {
-        CompletableFuture.allOf(
-            // delete remote data ids set in root network
-            studyServerExecutionService.runAsync(() -> reportService.deleteReports(rootNetworkInfos.stream().map(RootNetworkInfos::getReportUuid).toList())),
-            studyServerExecutionService.runAsync(() -> rootNetworkInfos.stream().map(rni -> rni.getNetworkInfos().getNetworkUuid()).filter(Objects::nonNull).forEach(equipmentInfosService::deleteEquipmentIndexes)),
-            studyServerExecutionService.runAsync(() -> rootNetworkInfos.stream().map(rni -> rni.getNetworkInfos().getNetworkUuid()).filter(Objects::nonNull).forEach(networkStoreService::deleteNetwork)),
-            studyServerExecutionService.runAsync(() -> rootNetworkInfos.stream().map(rni -> rni.getCaseInfos().getCaseUuid()).filter(Objects::nonNull).forEach(caseService::deleteCase))
+    public void deleteRootNetworkRemoteInfos(List<RootNetworkInfos> rootNetworkInfos, boolean deleteCase) {
+        // delete remote data ids set in root network
+        CompletableFuture.allOf(Stream.of(
```
flomillot: I know the code was before you, but what's the point of using futures and allOf if you don't use join or get after ?

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
```diff
@@ -235,18 +238,28 @@ public void deleteRootNetworks(StudyEntity studyEntity, Stream<UUID> rootNetwork
...
+                                studyServerExecutionService.runAsync(() ->
+                                        reportService.deleteReports(rootNetworkInfos.stream().map(RootNetworkInfos::getReportUuid).toList())),
+                                studyServerExecutionService.runAsync(() ->
+                                        rootNetworkInfos.stream().map(rni -> rni.getNetworkInfos().getNetworkUuid())
+                                                .filter(Objects::nonNull).forEach(equipmentInfosService::deleteEquipmentIndexes)),
+                                studyServerExecutionService.runAsync(() ->
+                                        rootNetworkInfos.stream().map(rni -> rni.getNetworkInfos().getNetworkUuid())
+                                                .filter(Objects::nonNull).forEach(networkStoreService::deleteNetwork)),
+                                !deleteCase ? null : studyServerExecutionService.runAsync(() ->
```
flomillot: It works, but it's a bit weird no ? And not very clear to read for me.
What about something like this ?

`List<CompletableFuture<?>> futures = new ArrayList<>();
futures.add(studyServerExecutionService.runAsync(() -> ...));
futures.add(studyServerExecutionService.runAsync(() -> ...));
if (deleteCase) {
    futures.add(studyServerExecutionService.runAsync(() -> ...));
}
CompletableFuture.allOf(futures.toArray(CompletableFuture[]::new));`

src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
Meklo: I thought it made more sense semantically to have a dedicated term and endpoint since it's not just an invalidation, wdyt @etiennehomer ?

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
Meklo: Yep I didn't question this bit of code but the parallelism is defeated here apparently

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
Meklo: Fair enough

src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
Meklo: Yep it invalidates all built nodes the same way as unbuildAllNodes endpoint plus it deletes all networks associated to root nodes as well as elasticssearch data and report nodes.
I think it covers all that can be temporaly deleted for space sake without actual data loss like modifications, parameters, etc..

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
Meklo: I've added the join but I need to retest it to check if it still works as intended

src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
flomillot: I agree. So you do more that what's described in the ticket, right ?

src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
flomillot: You say it deletes all associated networks. So why don't you call https://github.com/gridsuite/study-server/blob/2eeeccbcb72f66accabdb488c0ce1d8ce291a733/src/main/java/org/gridsuite/study/server/service/RootNetworkService.java#L240 directly ?
From what I understand you delete the network data but not the network itself.

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
flomillot: Okay, maybe check with the person who did it to be sure.

src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
etiennehomer: Before partial variants in the network-store, I would easily say that the unbuild of the root node is the same treatment as for the other nodes.
But now that we have partial variants, actually it's partly different. But not so much. The consequence is the delete of a network and its equipments in the network-store. 

To me, the main question is: do we want this action to be open to the front (can it be launch by the users) or we want to keep it as a supervision action ?

For now, I would rather say that it's fine to keep it as a supervision action.

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -3912,4 +3929,22 @@ private List<CurrentLimitViolationInfos> getCurrentLimitViolations(UUID nodeUuid
...
+
+    @Transactional
+    public void unmountStudyRootNetwork(UUID studyUuid, UUID rootNetworkUuid) {
+        StudyEntity study = getStudy(studyUuid);
+        RootNetworkEntity rootNetwork = rootNetworkService.getRootNetwork(rootNetworkUuid)
+                .orElseThrow(() -> new StudyException(NOT_FOUND, "Root network not found"));
+
+        var rootNode = networkModificationTreeService.getStudyRootNodeUuid(studyUuid);
+        try {
+            doUnbuildNodeTree(studyUuid, rootNode, true);
```
etiennehomer: you should rather delete other nodes first. Maybe it works like this. But if the code crashes, it's preferable to have : 
- only the root node built 
and not : 
- some modification nodes built but the root node unbuilt

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2685,6 +2694,7 @@ private UUID handleSensitivityAnalysisRequest(StudyEntity study, UUID nodeUuid,
 
         updateComputationResultUuid(nodeUuid, rootNetworkUuid, result, SENSITIVITY_ANALYSIS);
         notificationService.emitStudyChanged(study.getId(), nodeUuid, rootNetworkUuid, NotificationService.UPDATE_TYPE_SENSITIVITY_ANALYSIS_STATUS);
+        notificationService.emitElementUpdated(study.getId(), userId);
```
etiennehomer, line 2591: Validated by POs ? Do we want to change the actions that do emitElementUpdated() ?

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
Meklo: Apparently it was intentional to drop the .get() so maybe we should just remove CompletableFuture.allOf

etiennehomer: Hum to me, the join() does not ensure the paralllism but it makes the code wait for all the futures to finish before executing the next instructions. I would rather keep the previous behavior to make the execution faster

etiennehomer (changes requested):

src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
Meklo: We want to keep root networks entities to be able to rebuild them later, here we just want to delete the underlying data in network store server whereas deleteRootNetworks also deletes root network entity from study server if I'm not mistaken

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
Meklo: I've just check with @klesaulnier it seems alright to remove CompletableFuture.allOf altogether

src/main/java/org/gridsuite/study/server/service/StudyService.java
Meklo: Isn't it what is currently implemented ? Or I may be misunderstanding :

- doUnbuildNodeTree(studyUuid, rootNode, true); -> Unbuild all nodes 
- rootNetworkService.deleteRootNetworkRemoteInfos(List.of(rootNetwork.toDto()), false); -> Delete root node data 
- updateRootNetworkIndexationStatus(study, rootNetwork, RootNetworkIndexationStatus.NOT_INDEXED); -> Clean up following root node data deletion

src/main/java/org/gridsuite/study/server/service/StudyService.java
Meklo: It is part of the ticket description but maybe I should just make the inventory of actions that do update modification date first to crosscheck with PO which ones we do want to add to the pool

src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
flomillot: So actually, as said above, you don't delete the networks themself, rights ?
I just want to point out that, if I understand well, the endpoint description does not match with what it does exactly.

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -3912,4 +3929,22 @@ private List<CurrentLimitViolationInfos> getCurrentLimitViolations(UUID nodeUuid
...
             .toList();
     }
+
+    @Transactional
+    public void unmountStudyRootNetwork(UUID studyUuid, UUID rootNetworkUuid) {
+        StudyEntity study = getStudy(studyUuid);
+        RootNetworkEntity rootNetwork = rootNetworkService.getRootNetwork(rootNetworkUuid)
+                .orElseThrow(() -> new StudyException(NOT_FOUND, "Root network not found"));
+
+        var rootNode = networkModificationTreeService.getStudyRootNodeUuid(studyUuid);
```
etiennehomer:
```suggestion
        var rootNodeUuid = networkModificationTreeService.getStudyRootNodeUuid(studyUuid);
```

src/main/java/org/gridsuite/study/server/service/StudyService.java
etiennehomer: Yes perfect. I read it to quickly. Maybe add some comments as discussed

src/main/java/org/gridsuite/study/server/service/StudyService.java
etiennehomer: Discussed with PO: we'll enrich this list in another ticket. So to remove for now

src/test/java/org/gridsuite/study/server/SupervisionControllerTest.java
```diff
@@ -302,4 +302,24 @@ void testSupervisionStudiesBasicData() throws Exception {
...
+                .andExpect(status().isOk());
+
+        // Tree was unbuilt: variants dropped from network store
+        Mockito.verify(networkService, Mockito.times(1)).deleteVariants(eq(NETWORK_UUID), any());
+
+        // Remote root-network data was deleted, but the case was preserved (deleteCase = false)
+        Mockito.verify(rootNetworkService, Mockito.times(1))
+                .deleteRootNetworkRemoteInfos(any(), eq(false));
+
+        // Indexation flipped to NOT_INDEXED so the auto-detect path will reimport on reopen
```
etiennehomer, line 328: check ES data is empty:
assertIndexationCount(0, 0);

Meklo:
I've added a blocking mode for remote data deletion so that the load gets leaned and when can better mesure the job progress

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
```diff
@@ -253,6 +257,19 @@ public void deleteRootNetworkRemoteInfos(List<RootNetworkInfos> rootNetworkInfos
         rootNetworkNodeInfoService.deleteRootNetworkNodeRemoteInfos(rootNetworkInfos.stream().map(RootNetworkInfos::getRootNetworkNodeInfos).filter(Objects::nonNull).flatMap(Collection::stream).toList());
     }
 
+    public void invalidateRootNetworkRemoteInfos(List<RootNetworkInfos> rootNetworkInfos, boolean blocking) {
```
etiennehomer: You can merge this method with deleteRootNetworkRemoteInfos()
With a signature:
deleteRootNetworkRemoteInfos(List<RootNetworkInfos> rootNetworkInfos,boolean blocking, boolean deleteCase)

That's what you did before I think and it will work now that you have the skipDeleteVariant parameter in doUnbuildNodeTree()

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2080,16 +2086,16 @@ public void invalidateNodeTree(UUID studyUuid, UUID nodeUuid, UUID rootNetworkUu
     }
 
     private void invalidateNodeTree(UUID studyUuid, UUID nodeUuid, UUID rootNetworkUuid, InvalidateNodeTreeParameters invalidateTreeParameters) {
-        invalidateNodeTree(studyUuid, nodeUuid, rootNetworkUuid, invalidateTreeParameters, false);
+        invalidateNodeTree(studyUuid, nodeUuid, rootNetworkUuid, invalidateTreeParameters, false, false);
     }
 
-    public void invalidateNodeTree(UUID studyUuid, UUID nodeUuid, UUID rootNetworkUuid, InvalidateNodeTreeParameters invalidateTreeParameters, boolean blocking) {
+    public void invalidateNodeTree(UUID studyUuid, UUID nodeUuid, UUID rootNetworkUuid, InvalidateNodeTreeParameters invalidateTreeParameters, boolean blocking, boolean skipDeleteVariants) {
```
etiennehomer: It's not really linked to your PR. But this method is always called with blocking=false. We could remove it ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -3921,4 +3940,25 @@ private List<CurrentLimitViolationInfos> getCurrentLimitViolations(UUID nodeUuid
             .map(l -> new CurrentLimitViolationInfos(l.getSubjectId(), null))
             .toList();
     }
+
+    @Transactional
```
etiennehomer: To remove as it requests many other servers ?

etiennehomer (changes requested):

src/main/java/org/gridsuite/study/server/service/SupervisionService.java
```diff
@@ -365,6 +370,17 @@ public void unbuildAllNodes(UUID studyUuid) {
         LOGGER.trace("Nodes builds deletion for study {} in : {} seconds", studyUuid, TimeUnit.NANOSECONDS.toSeconds(System.nanoTime() - startTime.get()));
     }
 
+    @Transactional
```
etiennehomer: Same here ?

etiennehomer (changes requested):

src/main/java/org/gridsuite/study/server/service/StudyService.java
Meklo: Sounds legit

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
Meklo: I still need to differentiate ES index deletion and case deletion based on invalidation path but it works yes

Meklo: I had to keep them separate in the end since both path don't handle transactions the same way now

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
```diff
@@ -335,4 +353,31 @@ public boolean isRootNetworkTagExistsInStudy(UUID studyUuid, String rootNetworkT
         return rootNetworkRepository.findByTagAndStudyId(rootNetworkTag, studyUuid).isPresent() ||
                 rootNetworkRequestRepository.findByTagAndStudyUuid(rootNetworkTag, studyUuid).isPresent();
     }
+
+    @Transactional(readOnly = true)
```
etiennehomer: Not necessary ? To remove ?

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
```diff
@@ -335,4 +353,31 @@ public boolean isRootNetworkTagExistsInStudy(UUID studyUuid, String rootNetworkT
...
+    public List<RootNetworkEntity> getStudyRootNetwork(StudyEntity study) {
+        return study.getRootNetworks();
+    }
+
+    @Transactional(readOnly = true)
```
etiennehomer: Not necessary ? To remove ?

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
etiennehomer: Ok. Remove the blocking parameter if we do so

etiennehomer (changes requested):

etiennehomer (approved):

PR #979, opened 2026-04-17
report use uuid, parentOrder as key instead of having an extra v7 ordered uuid

jonenst:
"look at report-server/ the diffs from main I changed the PK from uuid to (uuid, order). This changes the API. adapt the code ok now look at study-server it needs updating for parentOrder"


jonenst:
Not planned for now, can come back to this if needed

PR #981, opened 2026-04-27
New notification on study creation

antoinebhs:
## PR Summary

Introduce a distinct notification on study creation to avoid race condition in directory server.
In the current code:
* directory server emits a notif on study creation
* study server emits a notif on study creation (resent by directory server if the study element is created in directory DB)
* study server emits a notif on study creation done (resent by directory server)
Sometimes we have 3 notifications (if the creation notif from study server arrives after the element is created in directory DB), sometimes 2 if not already created.

Instead of this system that leads to notification race conditions, I only rely on the notification of directory server and study creation finished. The study creation started notif is not consumed by directory server.
Relates to https://github.com/gridsuite/directory-server/pull/244


EtienneLt (approved):
code OK test OK

PR #982, opened 2026-04-28
Modify actions pool which update last modification date

Meklo:
## PR Summary

Add lastModificationDate update for following actions : 

- Update network vizualization parameters
- Add a new root network
- Update a root network associated case
- Delete a root network
- Recreate a study network from a case or root network
- Restore a deleted node
- Update the node aliases
- Create a dynamic simulation event
- Update a dynamic simulation event
- Delete a dynamic simulation event
- Launch any type of computation


AbdelHedhili (changes requested):
Can you add / modify the tests to check for the newly emitted msg

Meklo:
My bad, I think initially it didn't require test changes and I forgot to add them as the scope grew
I also added the launching of computations to the pool following the green flag of a PO

AbdelHedhili (approved):

PR #988, opened 2026-05-05
rename composite modif

souissimai:


src/main/java/org/gridsuite/study/server/dto/modification/NetworkModificationMetadata.java
```diff
@@ -29,4 +30,7 @@ public class NetworkModificationMetadata {
     String description;
     @Schema(description = "Modification type")
     String type;
+    @Schema(description = "Modification name (only present for composite modifications)")
+    @JsonInclude(JsonInclude.Include.NON_NULL)
+    String name;
```
Mathieu-Deharbe, line 35: To be checked with @SlimaneAmar .

It works fine and fits with the comment (`Metadata of the network modifications used in gridstudy tree view`) but the definition is mine and maybe those metadata are not supposed to have data specific to some kind of modifications.

Mathieu-Deharbe (approved):

PR #989, opened 2026-05-12
Fix : Wrong root network node maked as built after voltage init modifications

ayolab:
In insertVoltageInitModifications : Modification application contexts are filtered to the target root network but results were indexed against the full root network list, always attributing the built status to the first root network instead of the one voltage init was applied on.


SlimaneAmar (approved):

dbraquart (approved):
code: ok
tests: ok (with 2 and 3 root networks)

PR #990, opened 2026-05-12
Fix import parameters

etiennehomer:
## PR Summary
We now receive import parameters as Object from the network-conversion-server. 
So we serialize it before storing them in the database.
And when reading it from the database, we deserialize then to rebuild the object.
This avoids errors such as : 
Caused by: com.powsybl.network.conversion.server.NetworkConversionException: com.powsybl.commons.PowsyblException: Unexpected value for minimalValidationLevel:

Using a Map<String, Object> for importParameters in RootNetworkInfos allows us to remove the importParametersRaw field. importParameters field can be used for the communications with the front and powsybl-network-conversion-server.

Data migration: empty string is not used as a magic value anymore. They are replaced by the real param default value: null.

Linked PR:
https://github.com/powsybl/powsybl-network-conversion-server/pull/282
https://github.com/gridsuite/gridstudy-app/pull/3954


src/main/java/org/gridsuite/study/server/utils/JsonUtils.java
```diff
@@ -55,14 +62,50 @@ public static JsonNode nodeAt(@NonNull final JsonNode node, @NonNull final JsonP
...
+        }
+
+        rawParams.forEach((key, value) -> {
+            if (value == null) {
+                result.put(key, null);
+                return;
+            }
+            try {
+                result.put(key, MAPPER.readValue(value, Object.class));
+            } catch (JsonProcessingException e) {
```
AbdelHedhili:
```suggestion
            try {
                result.put(key, value == null ? null : MAPPER.readValue(value, Object.class));
            } catch (JsonProcessingException e) {
```

src/main/java/org/gridsuite/study/server/utils/JsonUtils.java
```diff
@@ -14,12 +14,19 @@
...
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.function.Predicate;
 
+import static org.gridsuite.study.server.error.StudyBusinessErrorCode.UNPROCESSABLE_IMPORT_PARAMETER;
+
 public final class JsonUtils {
+    private static final ObjectMapper MAPPER = new ObjectMapper();
```
AbdelHedhili: Why are you creating your own ObjectMapper ? Use the one spring provides or else you bypass all the configuration which can cause (and already have caused) issues.

src/main/java/org/gridsuite/study/server/utils/JsonUtils.java
```diff
@@ -55,14 +62,50 @@ public static JsonNode nodeAt(@NonNull final JsonNode node, @NonNull final JsonP
         return nodeAt(node, jsonNode -> !jsonNode.isMissingNode(), pointers);
     }
 
-    public static String getModificationContextJsonString(ObjectMapper objectMapper, Pair<String, List<ModificationApplicationContext>> modificationContextInfos) {
+    public static String getModificationContextJsonString(Pair<String, List<ModificationApplicationContext>> modificationContextInfos) {
         try {
-            ObjectNode modificationJson = (ObjectNode) objectMapper.readTree(modificationContextInfos.getFirst());
-            ObjectNode modificationContextJson = objectMapper.valueToTree(modificationContextInfos);
+            ObjectNode modificationJson = (ObjectNode) MAPPER.readTree(modificationContextInfos.getFirst());
+            ObjectNode modificationContextJson = MAPPER.valueToTree(modificationContextInfos);
```
AbdelHedhili: should be reverted

src/main/resources/db/changelog/db.changelog-master.yaml
```diff
@@ -374,3 +374,6 @@ databaseChangeLog:
   - include:
       file: changesets/changelog_20260319T184344Z.xml
       relativeToChangelogFile: true
+  - include:
+      file: changesets/changelog_20260519T112945Z.xml
+      relativeToChangelogFile: true
```
AbdelHedhili:
```suggestion
      relativeToChangelogFile: true

```

src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
```diff
@@ -149,7 +149,7 @@ public NetworkModificationsResult createModification(UUID groupUuid,
         HttpHeaders headers = new HttpHeaders();
         headers.setContentType(MediaType.APPLICATION_JSON);
 
-        HttpEntity<String> httpEntity = new HttpEntity<>(getModificationContextJsonString(objectMapper, modificationContextInfos), headers);
+        HttpEntity<String> httpEntity = new HttpEntity<>(getModificationContextJsonString(modificationContextInfos), headers);
```
AbdelHedhili: to revert

AbdelHedhili (changes requested):

src/main/java/org/gridsuite/study/server/utils/JsonUtils.java
etiennehomer: it was easier in term of code design. But yes, wrong idea. Fixed

src/main/resources/db/changelog/db.changelog-master.yaml
etiennehomer: done

src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
etiennehomer: done

src/main/java/org/gridsuite/study/server/utils/JsonUtils.java
etiennehomer: done

src/main/java/org/gridsuite/study/server/utils/JsonUtils.java
etiennehomer: done

AbdelHedhili (approved):

PR #996, opened 2026-05-21
refactor computations parameters

AbdelHedhili:


src/main/java/org/gridsuite/study/server/dto/computation/ComputationsParameters.java
```diff
@@ -0,0 +1,24 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package org.gridsuite.study.server.dto.computation;
+
+import lombok.Builder;
+
+import java.util.UUID;
+
+@Builder
+public record ComputationsParameters(UUID loadFlowParametersUuid,
```
flomillot:
```suggestion
public record ComputationParameterIDs(UUID loadFlowParametersUuid,
```
Or something like that.

src/main/java/org/gridsuite/study/server/service/common/ComputationParametersService.java
flomillot, line 1: Why not in the common directory ?

src/main/java/org/gridsuite/study/server/service/ComputationParametersService.java
```diff
@@ -0,0 +1,288 @@
...
+                        () -> shortCircuitService.createParameters(null),
+                        ComputationsParameters.ComputationsParametersBuilder::shortCircuitParametersUuid,
+                        ComputationsParameters::shortCircuitParametersUuid,
+                        "ShortCircuit analysis"),
+                new ComputationParametersDefinition(
+                        ComputationType.DYNAMIC_SIMULATION,
+                        StudyEntity::getDynamicSimulationParametersUuid,
+                        UserProfileInfos::getDynamicSimulationParameterId,
+                        dynamicSimulationService,
+                        dynamicSimulationService::createDefaultParameters,
```
flomillot: I think you can add this method to the new interface (so the same name for all services)

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -294,128 +267,16 @@ private void insertStudy(UUID studyUuid, String userId, NetworkInfos networkInfo
...
-        // no profile, or no/bad voltage init parameters in profile => use default values
-        try {
-            return voltageInitService.createVoltageInitParameters(null);
-        } catch (final Exception e) {
-            LOGGER.error("Error while creating default parameters for voltage init", e);
-            return null;
-        }
-    }
-
     private UUID createDefaultNetworkVisualizationParameters(String userId, UserProfileInfos userProfileInfos) {
```
flomillot, line 281: Not this one ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -572,26 +583,6 @@ private Optional<DeleteStudyInfos> doDeleteStudyIfNotCreationInProgress(UUID stu
...
-        if (uuid != null) {
-            try {
-                pccMinService.deletePccMinParameters(uuid);
-            } catch (Exception e) {
-                LOGGER.error("Could not delete pcc min parameters with uuid:" + uuid, e);
-            }
-        }
-    }
-
     private void removeNetworkVisualizationParameters(@Nullable UUID uuid) {
```
flomillot, line 574: Same

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -3666,6 +3411,16 @@ public boolean createOrUpdatePccMinParameters(StudyEntity studyEntity, String pa
         return userProfileIssue;
     }
 
+    private void removePccMinParameters(@Nullable UUID uuid) {
```
flomillot: Why is this one here ?

src/main/java/org/gridsuite/study/server/service/ComputationParametersService.java
flomillot: And also you already have the service in the object. So this line is redundant, you can access it from the service directly.

src/main/java/org/gridsuite/study/server/service/ComputationParametersService.java
```diff
@@ -0,0 +1,288 @@
...
+
+    public void deleteComputationParameters(UUID parametersUuid, ComputationParameters computationParameters, String computationType) {
+        if (parametersUuid != null) {
+            try {
+                computationParameters.deleteParameters(parametersUuid);
+            } catch (Exception e) {
+                LOGGER.error("Could not remove {} parameters with uuid: {}", computationType, parametersUuid, e);
+            }
+        }
+    }
```
flomillot: Do we really need this method ? It's not possible to slightly update the interface method and use it directly instead ?

flomillot: Can be 

```
public interface ComputationParameters {

    default void deleteParameters(UUID parametersUuid) {
        // Code commun exécuté pour toutes les implémentations
        // Exemple : log, validation, cleanup, audit, vérification, etc.

        doDeleteParameters(parametersUuid);

        // Éventuellement du code commun après la suppression
    }

    void doDeleteParameters(UUID parametersUuid);
}
```

and then in every class

```
@Override
public void doDeleteParameters(UUID parametersUuid) {
    // code spécifique à cette classe
}
```

src/main/java/org/gridsuite/study/server/service/ComputationParametersService.java
```diff
@@ -0,0 +1,288 @@
...
+
+        try {
+            return definition.defaultParametersSupplier().get();
+        } catch (Exception e) {
+            LOGGER.error("Error while creating default parameters for {}", definition.defaultErrorLabel(), e);
+            return null;
+        }
+    }
+
+    private UUID createDefaultStateEstimationParameters() {
```
flomillot: You could not include this one also ?

flomillot: Considering the code above, I think it should work.

src/main/java/org/gridsuite/study/server/service/common/ComputationParametersService.java
```diff
@@ -0,0 +1,288 @@
...
+        computationParametersDefinitions.forEach(definition ->
+                definition.parametersSetter().accept(parametersBuilder, createDefaultParameters(userId, userProfileInfos, definition)));
+        return parametersBuilder.build();
+    }
+
+    public ComputationsParameters duplicateParameters(StudyEntity sourceStudyEntity) {
+        ComputationsParameters.ComputationsParametersBuilder parametersBuilder = ComputationsParameters.builder();
+        computationParametersDefinitions.forEach(definition -> {
+            UUID sourceParametersUuid = definition.studyParameterGetter().apply(sourceStudyEntity);
+            if (sourceParametersUuid != null) {
```
flomillot, line 146: No exception or log if the source parameters UID is null ? It's OK to inject null UID ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -160,42 +165,43 @@ public enum ReportType {
...
-        StudyConfigService studyConfigService,
-        NadConfigService nadConfigService,
-        FilterService filterService,
-        StateEstimationService stateEstimationService,
-        PccMinService pccMinService,
-        @Lazy StudyService studyService,
-        RootNetworkService rootNetworkService,
-        RootNetworkNodeInfoService rootNetworkNodeInfoService,
-        DirectoryService directoryService) {
+            StudyRepository studyRepository,
```
flomillot: Bad indentation.

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -545,17 +552,21 @@ private Optional<DeleteStudyInfos> doDeleteStudyIfNotCreationInProgress(UUID stu
...
+                                .shortCircuitParametersUuid(s.getShortCircuitParametersUuid())
+                                .dynamicSimulationParametersUuid(s.getDynamicSimulationParametersUuid())
+                                .voltageInitParametersUuid(s.getVoltageInitParametersUuid())
+                                .securityAnalysisParametersUuid(s.getSecurityAnalysisParametersUuid())
+                                .sensitivityAnalysisParametersUuid(s.getSensitivityAnalysisParametersUuid())
+                                .dynamicSecurityAnalysisParametersUuid(s.getDynamicSecurityAnalysisParametersUuid())
+                                .dynamicMarginCalculationParametersUuid(s.getDynamicMarginCalculationParametersUuid())
+                                .stateEstimationParametersUuid(s.getStateEstimationParametersUuid())
+                                .pccMinParametersUuid(s.getPccMinParametersUuid())
+                                .build()
```
flomillot: Why all these lines here ? You can't do it directly in the new service ? You need one argument, the study, and then you can retrieve the ID in the service.

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -716,46 +701,6 @@ private StudyEntity duplicateStudy(BasicStudyInfos studyInfos, UUID sourceStudyU
...
-        if (sourceStudyEntity.getSensitivityAnalysisParametersUuid() != null) {
-            copiedSensitivityAnalysisParametersUuid = sensitivityAnalysisService.duplicateSensitivityAnalysisParameters(sourceStudyEntity.getSensitivityAnalysisParametersUuid());
-        }
-
-        UUID copiedVoltageInitParametersUuid = null;
-        if (sourceStudyEntity.getVoltageInitParametersUuid() != null) {
-            copiedVoltageInitParametersUuid = voltageInitService.duplicateVoltageInitParameters(sourceStudyEntity.getVoltageInitParametersUuid());
-        }
-
         UUID copiedNetworkVisualizationParametersUuid = null;
```
flomillot, line 692: Same remark as before.

src/main/java/org/gridsuite/study/server/service/ComputationParametersService.java
```diff
@@ -0,0 +1,288 @@
...
+                );
+            }
+        }
+
+        if (existingParametersUuid == null) {
+            UUID newParametersUuid = createParameters.apply(parameters);
+            studyParameterSetter.accept(studyEntity, newParametersUuid);
+        } else {
+            updateParameters.accept(existingParametersUuid, parameters);
+        }
```
flomillot: Duplicated code, why don't you call the method below ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2729,7 +2609,6 @@ private UUID handleShortCircuitRequest(StudyEntity studyEntity, UUID nodeUuid, U
         updateComputationResultUuid(nodeUuid, rootNetworkUuid, result, computationType);
         notificationService.emitStudyChanged(studyEntity.getId(), nodeUuid, rootNetworkUuid,
                 busId.isEmpty() ? NotificationService.UPDATE_TYPE_SHORT_CIRCUIT_STATUS : NotificationService.UPDATE_TYPE_ONE_BUS_SHORT_CIRCUIT_STATUS);
-        notificationService.emitElementUpdated(studyEntity.getId(), userId);
```
flomillot, line 2732: Is it normal to remove this notification ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -3590,7 +3340,6 @@ private UUID handlePccMinRequest(StudyEntity studyEntity, UUID nodeUuid, UUID ro
         UUID result = pccMinService.runPccMin(networkUuid, variantId, runPccMinParametersInfos, new ReportInfos(reportUuid, nodeUuid), receiver, userId);
         updateComputationResultUuid(nodeUuid, rootNetworkUuid, result, PCC_MIN);
         notificationService.emitStudyChanged(studyEntity.getId(), nodeUuid, rootNetworkUuid, NotificationService.UPDATE_TYPE_PCC_MIN_STATUS);
-        notificationService.emitElementUpdated(studyEntity.getId(), userId);
```
flomillot, line 3593: Same

src/main/java/org/gridsuite/study/server/dto/computation/ComputationsParameters.java
basseche: Agree you can even name it ComputationParameterUUIDs

basseche (approved):

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
AbdelHedhili: nothing to do with computations

src/main/java/org/gridsuite/study/server/service/StudyService.java
AbdelHedhili: This isn't a computation so it doesn't make sense to group them together

src/main/java/org/gridsuite/study/server/service/StudyService.java
AbdelHedhili: Good catch, it's absolutely not normal, that's a conflict blunder.

src/main/java/org/gridsuite/study/server/service/common/ComputationParametersService.java
AbdelHedhili: If you check the code before the refacto you'll see that I do exactly the same thing as before, if the source study doesn't have a parameterId for a given computation type I just don't try to duplicate this parameter. No log or anything, same behavior.

src/main/java/org/gridsuite/study/server/service/ComputationParametersService.java
AbdelHedhili: Not really, or I don't see how, the problem here is that the code is 100 % identical with all service except state estimation that has only the call to createDefault that is like the others. 

If I follow what you suggest here I'd find myself with basically one line that is common and the rest that I need to duplicate to all the services.

I instead removed this method and just added a check on the createDefaultParameters to skip the part that is not needed if the computationType is stateestim

flomillot: Well considering
https://github.com/gridsuite/study-server/blob/5339f2e06686ee0e3556e53ca2cc9b76cfb89e2e/src/main/java/org/gridsuite/study/server/service/common/ComputationParametersService.java#L132
and
image
I don't see why you needed to add code. 
Did I miss something ?

src/main/java/org/gridsuite/study/server/service/common/ComputationParametersService.java
flomillot: Yeah I know, and it's not supposed to happen because we are supposed to insert default parameters at the creation of a study, but if it does (the code change, or something I haven't noticed), it's nasty, no ? Shouldn't we at least create default parameters if we don't find existing parameters ? 
But if you don't want to change the existing behavior, it's okay as you want.

src/main/java/org/gridsuite/study/server/service/StudyService.java
flomillot: @coderabbitai Check if other notifications have been removed.

src/main/java/org/gridsuite/study/server/service/ComputationParametersService.java
AbdelHedhili: Nope you are right it's useless

src/main/java/org/gridsuite/study/server/service/ComputationParametersService.java
flomillot: I didn't notice at first but you can do exactly the same with `createDefaultParameters`

flomillot (approved):

PR #1004, opened 2026-06-04
transmit CompositesToBeInserted to network modifications server

Mathieu-Deharbe:


src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -691,27 +691,27 @@ public ResponseEntity<Void> moveOrCopyModifications(@PathVariable("studyUuid") U
...
     }
 
-    private void handleInsertCompositeNetworkModifications(UUID targetStudyUuid, UUID targetNodeUuid, List<Pair<UUID, String>> modificationsToCopy, String userId, CompositeModificationsActionType action) {
+    private void handleInsertCompositeNetworkModifications(UUID targetStudyUuid, UUID targetNodeUuid, List<Object> compositesToBeInserted, String userId, CompositeModificationsActionType action) {
         studyService.assertNoBlockedNodeInStudy(targetStudyUuid, targetNodeUuid);
         studyService.invalidateNodeTreeWithLF(targetStudyUuid, targetNodeUuid);
         try {
-            studyService.insertCompositeNetworkModifications(targetStudyUuid, targetNodeUuid, modificationsToCopy, userId, action);
+            studyService.insertCompositeNetworkModifications(targetStudyUuid, targetNodeUuid, compositesToBeInserted, userId, action);
         } finally {
```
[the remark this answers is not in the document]
Mathieu-Deharbe: I use `List<Object>` in order to avoid the creation of a dto in study-server because it is not really used in that server and would require upkeep. But it is validated in the next called microservices (network-modification-server) which will use the data.

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
```diff
@@ -99,13 +102,60 @@ public boolean elementExists(UUID directoryUuid, String elementName, String type
         return response.getStatusCode() == HttpStatus.OK;
     }
 
+    /**
+     *
+     * @param elementsUuids all the element uuids of the shared composites that need to be referenced in directory server
+     * @param userId id of the user who started the insertion
+     * @param targetNodeUuid where the new references will point
+     */
+    public void addReferencesToSharedComposites(@NonNull List<UUID> elementsUuids, String userId, UUID targetNodeUuid) {
```
flomillot: The Java doc and method name is not clear for me. What is the element UID of a shared composite ? What is an insertion ? The references will point to what ? We add or create a new reference ?

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
```diff
@@ -99,13 +102,60 @@ public boolean elementExists(UUID directoryUuid, String elementName, String type
...
+
+            ReferenceAttributes referenceAttributes = new ReferenceAttributes(targetNodeUuid, STUDY_NODE);
+
+            HttpEntity<ReferenceAttributes> requestEntity = new HttpEntity<>(referenceAttributes, headers);
+            restTemplate.exchange(getDirectoryServerServerBaseUri() + path, HttpMethod.POST, requestEntity, ElementAttributes.class);
+        });
+    }
+
+    /**
+     * remove references from the shared modification in directory server
```
flomillot: Of or from ? What kind of references ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2066,7 +2094,16 @@ public void restoreNetworkModifications(UUID studyUuid, UUID nodeUuid, List<UUID
                 throw new StudyException(NOT_ALLOWED);
             }
             UUID groupId = networkModificationTreeService.getModificationGroupUuid(nodeUuid);
+            Map<UUID, UUID> referenceToBeRecreated = networkModificationService.getReferencesData(modificationsUuids);
```
flomillot: You fetch the references before restoring the modifications. The stashed modification are filtered. References to be recreated will be empty. You should restore them before fetching the references. Like it s done for restoreNodes. Am I right ? 
Also maybe add a test for this.

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
```diff
@@ -99,13 +102,60 @@ public boolean elementExists(UUID directoryUuid, String elementName, String type
...
     }
 
+    /**
+     *
+     * @param elementsUuids all the element uuids of the shared composites that need to be referenced in directory server
+     * @param userId id of the user who started the insertion
+     * @param targetNodeUuid where the new references will point
+     */
+    public void addReferencesToSharedComposites(@NonNull List<UUID> elementsUuids, String userId, UUID targetNodeUuid) {
+        elementsUuids.forEach(elementUuid -> {
```
flomillot, line 114: The fact that the method do several rest call is not intuitive

src/main/java/org/gridsuite/study/server/service/StudyService.java
flomillot: The references already exist, right ? So we don't we create them, we add them ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2066,7 +2094,16 @@ public void restoreNetworkModifications(UUID studyUuid, UUID nodeUuid, List<UUID
...
             networkModificationService.restoreModifications(groupId, modificationsUuids);
+            // if there are references modifications in the unstashed netmods, those references had been removed and must be recreated in directory server
+            if (!referenceToBeRecreated.isEmpty()) {
```
flomillot: Maybe the question is dumb, but It's not dangerous to manage the stashed modification like that ? What is the real problem of the directory service pointing to stashed modification ?

Mathieu-Deharbe: This behavior has been removed after today's discussion : https://github.com/gridsuite/study-server/pull/1004/changes/585580691288d210df7ea5701599c155ab3b30fb

src/main/java/org/gridsuite/study/server/service/StudyService.java
Mathieu-Deharbe: Actually the references in the directory-server database had been deleted and really had to be recreated.

But it doesn't matter anymore, this behavior has been removed : https://github.com/gridsuite/study-server/pull/1004/changes/585580691288d210df7ea5701599c155ab3b30fb

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
Mathieu-Deharbe: I'd rather have an endpoint which accepts uuid's lists and do only one call but it doesn't. 

What would you prefer ? One call in the function and this function called for each elementsUuid ?

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
Mathieu-Deharbe: **From** : the shared modification in directory server owns a list of references and we remove one from it.

I think that the full name of the reference type would be: **_a reference to a use of this shared modification_**.

src/main/java/org/gridsuite/study/server/service/StudyService.java
Mathieu-Deharbe: Yep you might be right. Fortunately this behavior has been removed after today's discussion : https://github.com/gridsuite/study-server/pull/1004/changes/585580691288d210df7ea5701599c155ab3b30fb

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
Mathieu-Deharbe: Is this better : https://github.com/gridsuite/study-server/pull/1004/changes/8a9e97038fe00a59a0f4b0243f3871362ba54e62

souissimai (approved):

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
flomillot: The confusion for me is between the technical word 'reference', the object we store in directory-server, and a modification of type reference.
Its not the same object in network-modification and directory-server.
https://github.com/gridsuite/directory-server/blob/324c4ab61f0b0776eef8ab8749ddea021826a7f7/src/main/java/org/gridsuite/directory/server/DirectoryService.java#L341
In the second one, we create a reference to... a modification reference ? Its confusing for me and not clear when we dont know the whole picture.

To recap : 
We have CompositeModification https://github.com/gridsuite/network-modification-server/blob/658bcb3d5c03caa86f22163d73a2ae3b167c991d/src/main/java/org/gridsuite/modification/server/entities/CompositeModificationEntity.java#L29
We have ReferenceModification https://github.com/gridsuite/network-modification-server/blob/56edd5d91147ebe6592bc4068c49083fb1dfe026/src/main/java/org/gridsuite/modification/server/entities/ModificationReferenceEntity.java#L25
Tbh I'm really confused and dont even know we add a reference of what to what.
Btw 'Creates...to...'is not correct in english I think, its either ' add to' or 'create for/of'.

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
flomillot: As discussed the other day, and doing REST calls with lot of composites will often happen, maybe update directory-server to accept several `ReferenceAttributes` to be coherent with other endpoints ?

Mathieu-Deharbe: Actually this is how I wanted those endpoints to be done initially (with arrays). But this is a bit late for this.

What do you think @SlimaneAmar ?

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
Mathieu-Deharbe: I agree. : Adding `references` to `modification references` is confusing. I think reference is a word too generic and will bring confusion. Add to this the fact that references don't really point to the modification references but to its containers. 
But it is hard to come up with a good name approved by all.

The "to" doesnt refer to "create". That would indeed be bad english. It refers to "references". Some References **To** Shared Composites

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
Mathieu-Deharbe: TODO added : https://github.com/gridsuite/study-server/pull/1004/changes/d6b26e8ddffef4a21f26a53cbdfb6fc5df30e42c

flomillot (approved):
Code OK (not tested)

PR #1007, opened 2026-06-10
Fix invalidated multiroot studies blocked nodes

Meklo:
## PR Summary

Unblock node tree only after all root networks have been invalidated. The previous approach oversaw that doUnbuildNodeTree locks all root networks so after unlocking the first root network it was getting locked again one the second pass


etiennehomer (changes requested):
We should execute testInvalidateStudy() with 2 root networks and check that all nodes are unblocked

etiennehomer (approved):

PR #1010, opened 2026-06-16
Clean tests instability

Meklo:
## PR Summary

Two distinct causes of instability were identified and fixed.

### 1. Races on non-blocking async deletions

Several cleanup paths dispatch remote-resource deletions as fire-and-forget tasks via `studyServerExecutionService.runAsync` (e.g. `invalidateRootNetworkRemoteInfos` with `blocking=false`). Tests that verify those deletions race the executor pool: the `verify` can run before the task does. Since `deleteCase` is the last task submitted, it's the one that fails most often (seen in `testCreateRootNetworkConsumerWithoutRequest`, and latent in `testDeleteRootNetwork`).

**Fix:** use `synchronizeStudyServerExecutionService` when needed

### 2. Double deletion of the creation request (`testCreateStudyWithErrorDuringCaseImport`)

On a synchronous 5xx during import, `deleteStudyIfNotCreationInProgress` is called concurrently by two paths:

- the `catch` block in `createStudy` (request thread)
- `consumeCaseImportFailed` (RabbitMQ consumer)

Depending on ordering, the consumer can read an already-deleted `study_creation_request`, fall into the branch that calls `getStudy` → `Study not found`, the exception is swallowed, and the `STUDY_CREATION_FINISHED` notification is never emitted → the test waits for a message that never arrives.

**Fix:** `doDeleteStudyIfNotCreationInProgress` made idempotent — if neither the creation request nor the study exists, it no-ops instead of throwing. The consumer then always reaches `emitStudyCreationError`, regardless of which path wins the race.


src/test/java/org/gridsuite/study/server/utils/TestUtils.java
```diff
@@ -307,4 +307,17 @@ public static void checkUpdateTypeMessageReceived(UUID studyUuid, UUID nodeUuid,
...
         ALL_COMPUTATION_STATUS.forEach(computationStatus -> checkUpdateTypeMessageReceived(studyUuid, nodeUuid, computationStatus, output, STUDY_UPDATE_DESTINATION));
     }
+
+    /**
+     * Forces {@link StudyServerExecutionService#runAsync} to execute its task inline on the calling
+     * thread instead of dispatching it to the executor pool. Used to make testing more consistent when
+     * functions launch new threads in non blocking mode
+     */
+
+    public static void executeAsyncOnMainThread(StudyServerExecutionService studyServerExecutionService) {
```
SlimaneAmar: Same as synchronizeStudyServerExecutionService ?

Meklo: Seems like it ! I'll try swaping it with yours

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -546,23 +546,22 @@ private Optional<DeleteStudyInfos> doDeleteStudyIfNotCreationInProgress(UUID stu
...
+            studyInfosService.deleteByUuid(studyUuid);
+            computationParametersService.deleteComputationsParameters(s);
+            removeNetworkVisualizationParameters(s.getNetworkVisualizationParametersUuid());
+            removeSpreadsheetConfigCollection(s.getSpreadsheetConfigCollectionUuid());
+            removeWorkspacesConfig(s.getWorkspacesConfigUuid());
+            removeNadConfigs(s.getNadConfigsUuids().stream().toList());
             deleteStudyInfos = new DeleteStudyInfos(rootNetworkInfos, modificationGroupUuids);
         } else {
-            studyCreationRequestRepository.deleteById(studyCreationRequestEntity.get().getId());
+            studyCreationRequestEntity.ifPresent(creationRequestEntity -> studyCreationRequestRepository.deleteAllByIdInBatch(List.of(creationRequestEntity.getId())));
```
SlimaneAmar: Why `deleteAllByIdInBatch` ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -546,23 +546,22 @@ private Optional<DeleteStudyInfos> doDeleteStudyIfNotCreationInProgress(UUID stu
...
-                networkModificationTreeService.doDeleteTree(studyUuid);
-                studyRepository.deleteById(studyUuid);
-                studyInfosService.deleteByUuid(studyUuid);
-                computationParametersService.deleteComputationsParameters(s);
-                removeNetworkVisualizationParameters(s.getNetworkVisualizationParametersUuid());
-                removeSpreadsheetConfigCollection(s.getSpreadsheetConfigCollectionUuid());
-                removeWorkspacesConfig(s.getWorkspacesConfigUuid());
-                removeNadConfigs(s.getNadConfigsUuids().stream().toList());
-            });
+            StudyEntity s = studyEntity.get();
```
SlimaneAmar, line 553: No need to delete something ih study is absent ?

SlimaneAmar (approved):

PR #1013, opened 2026-07-06
feat: add computation quotas check

GhilesA:
## PR Summary

- send error when max quota is reached
- expose quota check status (can be enabled if necessary, default quota check is off for now)


src/test/java/org/gridsuite/study/server/dto/OperationTypeTest.java
```diff
@@ -0,0 +1,29 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package org.gridsuite.study.server.dto;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
```
etiennehomer: author

src/test/java/org/gridsuite/study/server/service/UserAdminServiceTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.ArgumentMatchers.isNull;
+import static org.mockito.ArgumentMatchers.matches;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
```
etiennehomer, line 33: author

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -69,6 +69,7 @@ public class ConsumerService {
     private final RootNetworkNodeInfoService rootNetworkNodeInfoService;
     private final DirectoryService directoryService;
     private final ComputationParametersService computationParametersService;
+    private final UserAdminService userAdminService;
 
     public ConsumerService(ObjectMapper objectMapper,
```
etiennehomer, line 74: no indentation changes in the class plz

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -463,47 +478,59 @@ private void handleUnblockNode(NodeReceiver receiverObj, ComputationType computa
...
+
+                    String errorMessage = (String) msg.getHeaders().get(HEADER_ERROR_MESSAGE);
+                    String userId = (String) msg.getHeaders().get(HEADER_USER_ID);
 
-                String errorMessage = (String) msg.getHeaders().get(HEADER_ERROR_MESSAGE);
-                String userId = (String) msg.getHeaders().get(HEADER_USER_ID);
-                notificationService.emitComputationDebugFileStatus(studyUuid, receiverObj.getNodeUuid(), receiverObj.getRootNetworkUuid(), computationType, userId, resultUuid, errorMessage);
-            }));
+                    // free quota
+                    if (userId != null) {
```
etiennehomer: In debug mode, at the end of the computation, you'll receive 2 notifications :
- the classic one that says that the computation is done
- the one that said that the debug zip is ready for download.
So you should not `endOperationWithQuota()` in consumeCalculationDebug(). That would free twice

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -165,6 +166,9 @@ public enum ReportType {
 
     private final StudyService self;
 
+    @Value("${study.enable-operation-quotas}")
+    private boolean shouldCheckOperationQuotas = true;
```
etiennehomer: The default value in the application.yaml is enough ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1009,6 +1013,7 @@ private void handleLoadflowRequest(StudyEntity studyEntity, UUID nodeUuid, UUID
                 new LoadFlowService.ParametersInfos(lfParametersUuid, withRatioTapChangers, isSecurityNode), lfReportUuid, userId);
         rootNetworkNodeInfoService.updateLoadflowResultUuid(nodeUuid, rootNetworkUuid, result, withRatioTapChangers);
 
+        userAdminService.startOperationWithQuota(userId, OperationType.mapFromComputationType(LOAD_FLOW), result);
```
etiennehomer: we keep the startOperationWithQuota() in StudyService and not LoadflowService.runLoadflow() ?
This would allow to write some code that call runLoadflow() and do not call startOperationWithQuota()
But maybe StudyService is fine

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -3734,4 +3771,26 @@ public void invalidateStudyRootNetwork(UUID studyUuid, UUID rootNetworkUuid, Str
         rootNetworkService.updateRootNetworkIndexationStatus(studyUuid, rootNetworkUuid, RootNetworkIndexationStatus.NOT_INDEXED);
         notificationService.emitRootNetworksUpdated(studyUuid);
     }
+
+    public void assertOnQuotasAvailability(ComputationType computationType, String userId) {
```
etiennehomer, line 3815: This method could be just a call to a user-admin-server endpoint that returns true/false (computation allowed or not)
But you did this in anticipation of the coming front giving the quota information to the users ? Makes sense in that case

etiennehomer (dismissed):
Not enough time to test and to review UT
But i'm ok with the global design

src/main/java/org/gridsuite/study/server/dto/OperationType.java
```diff
@@ -0,0 +1,41 @@
...
+    SHORT_CIRCUIT,
+    VOLTAGE_INIT,
+    PCC_MIN,
+    STATE_ESTIMATION,
+    BALANCE_ADJUSTEMENT,
+    DYNAMIC_SIMULATION,
+    DYNAMIC_SECURITY,
+    DYNAMIC_MARGIN;
+
+    public static OperationType mapFromComputationType(ComputationType computationType) {
```
etiennehomer: keep same enum labels between ComputationType and OperationType

src/main/java/org/gridsuite/study/server/dto/OperationType.java
```diff
@@ -0,0 +1,41 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package org.gridsuite.study.server.dto;
+
+/**
+ * @author Ghiles Abdellah {@literal <ghiles.abdellah at rte-france.com>}
+ */
+public enum OperationType {
```
etiennehomer: QuotaType ? Another name ?
Because cases and builds are not operation. They static stuff staying in the system

etiennehomer: And is it ok to use the same enum for elements that have quotas but different nature (number of LF at the same time VS number of cases stored in the system)

src/main/java/org/gridsuite/study/server/service/StudyService.java
GhilesA: Since the current StudyService component seems to be an orchestrator it's probably better to keep this logic in it and centralize the quota operation handling for now

src/main/java/org/gridsuite/study/server/service/StudyService.java
GhilesA: Exactly, but for security purpose we can move this function in user-admin-server. 
For now the logic is in study server since it's the one that is responsible for all the start of computations no need to expose it through user-admin-server

src/main/java/org/gridsuite/study/server/dto/OperationType.java
GhilesA: renamed to QuotaType
For the second question, in the base logic of this dev everything can have a max quota value even if the way to compute the current usage is not the same

EtienneLt (approved):

PR #1015, opened 2026-07-08
add asymmetrical load computation

EtienneLt:


souissimai (approved):
check duplicated code but not sure that you can manage

PR #1020, opened 2026-07-17
feat: add an endpoint returning the name and study of given nodes

flomillot:
Add `POST /v1/nodes/infos`, taking a list of node uuids and returning the name of each node and the study it belongs to. Unknown node uuids are omitted from the response.


src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -1554,6 +1554,17 @@ public ResponseEntity<List<ModificationsSearchResultByNode>> searchModifications
                 .body(studyService.searchModifications(rootNetworkUuid, userInput));
     }
 
+    @PostMapping(value = "/nodes/infos", produces = MediaType.APPLICATION_JSON_VALUE)
```
Meklo: Should be GET request no ? Or do we definitively need a body attached ?

src/main/java/org/gridsuite/study/server/service/NetworkModificationTreeService.java
```diff
@@ -115,6 +115,14 @@ public List<ModificationsSearchResultByNode> getNetworkModificationsByNodeInfos(
                 .toList();
     }
 
+    @Transactional(readOnly = true)
+    public List<NodeInfos> getNodesInfos(List<UUID> nodeUuids) {
+        // unknown node uuids are simply absent from the result
+        return networkModificationNodeInfoRepository.findAllByIdIn(nodeUuids).stream()
+                .map(nodeInfo -> new NodeInfos(nodeInfo.getId(), nodeInfo.getName(), nodeInfo.getNode().getStudy().getId()))
```
Meklo, line 122: N+1 here on getNode and getStudy. Is preloading them an option here or the workload is not significant enough to optimize it, wdyt ?

flomillot: I'm not aware of all the JPA optimization, how would you do it ?
@coderabbitai is this transactional method bad implemented ?

src/main/java/org/gridsuite/study/server/controller/StudyController.java
flomillot: Well UUIDs can be long and it's a list so I though we should use a body. We use both currently in the server.

flomillot: Ok I checked and actually the request param is the most used, I change it

src/main/java/org/gridsuite/study/server/service/NetworkModificationTreeService.java
Meklo: smth like this in the repo would preinitialize study data in one query :
```
    @EntityGraph(attributePaths = {"node", "node.study"})
    @Query(value = "SELECT n FROM NetworkModificationNodeInfoEntity n WHERE n.idNode IN (?1) ORDER BY n.columnPosition")
    List<NetworkModificationNodeInfoEntity> findAllByIdInWithNodeAndStudy(List<UUID> uuids);
```
    
@coderabbitai, it would be simpler than dto projection no ?

flomillot: Okay, thank you. I apply your suggestion but I just remove the node.study in the entity graph because the study ID is already included in the node entity as a foreign key. 
So it's not necessary and counterproductive to include the whole study entity in the loading.

src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -1554,6 +1554,17 @@ public ResponseEntity<List<ModificationsSearchResultByNode>> searchModifications
...
 
+    @GetMapping(value = "/nodes/infos", produces = MediaType.APPLICATION_JSON_VALUE)
+    @Operation(summary = "Get the name and the study of each given node")
+    @ApiResponses(value = {
+        @ApiResponse(responseCode = "200", description = "Infos of the found nodes, unknown nodes are omitted"),
+    })
+    public ResponseEntity<List<NodeInfos>> getNodesInfos(
+            @Parameter(description = "Node UUIDs") @RequestParam("ids") List<UUID> nodeUuids) {
+        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON)
+                .body(networkModificationTreeService.getNodesInfos(nodeUuids));
```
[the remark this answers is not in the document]
flomillot: There is no such a mechanism in study-server for reading data.

ghazwarhili (approved):

PR #1024, opened 2026-07-21
Entry endpoints

klesaulnier:
## PR Summary

Implements all entry endpoints to make all API calls from front pass through entry server rather than directly to target server


antoinebhs:
will probably clash with https://github.com/gridsuite/study-server/pull/1019/changes

src/main/java/org/gridsuite/study/server/controller/DirectoryController.java
```diff
@@ -0,0 +1,47 @@
...
+@RestController
+@RequestMapping(value = "/" + StudyApi.API_VERSION + "/directory")
+public class DirectoryController {
+    private final DirectoryService directoryService;
+
+    public DirectoryController(DirectoryService directoryService) {
+        this.directoryService = directoryService;
+    }
+
+    @GetMapping(value = "/elements", produces = MediaType.APPLICATION_JSON_VALUE)
```
antoinebhs, line 33: I wonder how to handle that: on main branch these similar endpoints are in StudyController.  
Maybe we shouodl create an issue to align with that?
To discuss with @SlimaneAmar

antoinebhs (commented):
need to add new servers to application-local?

PR #1025, opened 2026-07-21
refactor(RemoteServicesInspector): simplify getServicesInfo using List and plain loop

freddidierRTE:
## PR Summary

- Replace the raw CompletableFuture array with a List<CompletableFuture> to avoid the unchecked cast workaround and the need for the @SuppressWarnings annotation.

- The parallel stream was replaced by a sequential stream since parallelism is already handled by the async CompletableFuture calls themselves; using parallelStream() on the mapping step brought no real benefit here.

- The AtomicBoolean used to track partial results across a lambda is replaced by a plain boolean, improving readability now that the result collection is done in a standard for-loop.

- The PartialResultException is now thrown with the result HashMap directly, removing a redundant defensive copy (new HashMap<>(result)).

- Also fix a minor logging issue: replace string concatenation in the LOGGER.debug call with a parameterized message pattern {} to align with SLF4J best practices and avoid unnecessary string construction when debug logging is disabled.efacto getServiceInfo

### DIFF
--- a/src/main/java/org/gridsuite/study/server/service/RemoteServicesInspector.java
+++ b/src/main/java/org/gridsuite/study/server/service/RemoteServicesInspector.java
@@ -40,7 +40,6 @@ import java.time.Duration;
 import java.util.*;
 import java.util.Map.Entry;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.atomic.AtomicBoolean;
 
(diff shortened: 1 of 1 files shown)


jonenst (approved):

jonenst:
code OK
test OK

PR #1027, opened 2026-07-22
fix(ConsumerService): ensure study creation error notification is always emitted on case import failure

freddidierRTE:
Isolate deleteStudyIfNotCreationInProgress in its own try/catch in
   consumeCaseImportFailed so that emitStudyCreationError is always called
   even if the deletion fails due to a concurrent transaction that already
   cleaned up the StudyCreationRequestEntity.
   
   This was causing a flaky test (testCreateStudyWithErrorDuringCaseImport)
   when running the full test suite because the WireMock stub stubImportNetworkWithServerError
   simulates a situation that does not exist in production:
   
   it returns a 5xx HTTP response AND sends a DLX message simultaneously.
   
   In reality, a synchronous 5xx means the network conversion server
   never processed the request and therefore never publishes on case.import.start,
   so no DLX message would ever be generated.
   
   This artificial combination caused the catch block in
   createStudy and the DLX consumer to race against each other trying to
   delete the same StudyCreationRequestEntity. Depending on thread
   scheduling, the consumer's delete could throw an
   ObjectOptimisticLockingFailureException, which was silently caught at the
   outer level, preventing emitStudyCreationError from being called and
   leaving the test waiting for a message that never arrived.

**Solution was just to  catch an exception but as sonar complains with nested try/catch and complexity the method has been cut in two**

### DIFF
--- a/src/main/java/org/gridsuite/study/server/service/ConsumerService.java
+++ b/src/main/java/org/gridsuite/study/server/service/ConsumerService.java
@@ -344,30 +344,36 @@ public class ConsumerService {
         return message -> {
             String receiverString = message.getHeaders().get(HEADER_RECEIVER, String.class);
             String errorMessage = message.getHeaders().get(StudyConstants.HEADER_ERROR_MESSAGE, String.class);
-
             if (receiverString != null) {
-                CaseImportReceiver receiver;
                 try {
-                    receiver = objectMapper.readValue(URLDecoder.decode(receiverString, StandardCharsets.UTF_8),
-                        CaseImportReceiver.class);
-                    UUID studyUuid = receiver.getStudyUuid();
-                    String userId = receiver.getUserId();
-                    UUID rootNetworkUuid = receiver.getRootNetworkUuid();
-
-                    if (receiver.getCaseImportAction() == CaseImportAction.STUDY_CREATION) {
-                        studyService.deleteStudyIfNotCreationInProgress(studyUuid, userId);
-                        notificationService.emitStudyCreationError(studyUuid, userId, errorMessage);
-                    } else {
-                        if (receiver.getCaseImportAction() == CaseImportAction.ROOT_NETWORK_CREATION) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -344,32 +344,38 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
+        CaseImportReceiver receiver = objectMapper.readValue(URLDecoder.decode(receiverString, StandardCharsets.UTF_8),
+            CaseImportReceiver.class);
+        UUID studyUuid = receiver.getStudyUuid();
+        String userId = receiver.getUserId();
+        UUID rootNetworkUuid = receiver.getRootNetworkUuid();
+
+        if (receiver.getCaseImportAction() == CaseImportAction.STUDY_CREATION) {
+            try {
+                studyService.deleteStudyIfNotCreationInProgress(studyUuid, userId);
+            } catch (Exception e) {
```
jonenst, line 367: maybe catch ObjectOptimisticLockingFailureException instead and add comment that if there are races to delete the study we swaller the exception ?

src/main/java/org/gridsuite/study/server/service/ConsumerService.java
```diff
@@ -344,32 +344,38 @@ public Consumer<Message<String>> consumeCaseImportFailed() {
...
+                LOGGER.warn("Could not delete study during case import failure handling: {}", e.getMessage());
```
jonenst, line 368: maybe add the same handling or at least a comment on the other places in the code participating in this race ? If I understand correctly, StudyService::createStudy try catch ? To me even if it's not happening right now, we could some day be in a situation where the call to network-conversion-server to return 500 even though it published a message (maybe proxies, maybe the code has changed and it throws after publishing the message, maybe something else)

freddidierRTE:
replace by https://github.com/gridsuite/study-server/pull/1030

PR #1030, opened 2026-07-24
Fix race condition when deleting a study

freddidierRTE:
Resolved a race condition that caused testCreateStudyWithErrorDuringCaseImport to fail intermittently.

### DIFF
--- a/src/main/java/org/gridsuite/study/server/service/StudyService.java
+++ b/src/main/java/org/gridsuite/study/server/service/StudyService.java
@@ -67,6 +67,7 @@ import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort;
 import org.springframework.data.util.Pair;
 import org.springframework.lang.Nullable;
+import org.springframework.orm.ObjectOptimisticLockingFailureException;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.web.util.UriUtils;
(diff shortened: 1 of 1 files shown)


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -551,25 +552,30 @@ private Optional<DeleteStudyInfos> doDeleteStudyIfNotCreationInProgress(UUID stu
...
+                removeNetworkVisualizationParameters(s.getNetworkVisualizationParametersUuid());
+                removeSpreadsheetConfigCollection(s.getSpreadsheetConfigCollectionUuid());
+                removeWorkspacesConfig(s.getWorkspacesConfigUuid());
+                removeNadConfigs(s.getNadConfigsUuids().stream().toList());
+                deleteStudyInfos = new DeleteStudyInfos(rootNetworkInfos, modificationGroupUuidsNodeUuids);
+            } else {
+                studyCreationRequestEntity.ifPresent(creationRequestEntity -> studyCreationRequestRepository.deleteById(creationRequestEntity.getId()));
+            }
+        } catch (ObjectOptimisticLockingFailureException e) {
+            // This exception my be raise with a race condition when two threads try to delete the same study at the same time.
```
jonenst:
```suggestion
            // This exception may be raised with a race condition when two threads try to delete the same study at the same time. Instead of redesigning the whole system, we accept this design
```

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -551,25 +552,30 @@ private Optional<DeleteStudyInfos> doDeleteStudyIfNotCreationInProgress(UUID stu
...
+                removeSpreadsheetConfigCollection(s.getSpreadsheetConfigCollectionUuid());
+                removeWorkspacesConfig(s.getWorkspacesConfigUuid());
+                removeNadConfigs(s.getNadConfigsUuids().stream().toList());
+                deleteStudyInfos = new DeleteStudyInfos(rootNetworkInfos, modificationGroupUuidsNodeUuids);
+            } else {
+                studyCreationRequestEntity.ifPresent(creationRequestEntity -> studyCreationRequestRepository.deleteById(creationRequestEntity.getId()));
+            }
+        } catch (ObjectOptimisticLockingFailureException e) {
+            // This exception my be raise with a race condition when two threads try to delete the same study at the same time. 
+            LOGGER.error("Could not delete study with uuid:" + studyUuid, e);
```
jonenst:
```suggestion
            LOGGER.warn("Could not delete (maybe already deleted) study with uuid " + studyUuid, e);
```

jonenst (approved):

PR #1044, opened 2026-08-03
Entry endpoint from frontends

klesaulnier:
## PR Summary
Add endpoints in order to make all frontend pass through secured endpoints instead of reaching unsecured enpoints directly


src/main/java/org/gridsuite/study/server/controller/dynamicmargincalculation/DynamicMarginCalculationParametersController.java
```diff
@@ -0,0 +1,47 @@
...
+        this.dynamicMarginCalculationService = dynamicMarginCalculationService;
+    }
+
+    @GetMapping(value = "/providers", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<String> getProviders() {
+        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(dynamicMarginCalculationService.getProviders());
+    }
+
+    @GetMapping(value = "/parameters/{parameterUuid}", produces = MediaType.APPLICATION_JSON_VALUE)
+    public ResponseEntity<String> getParameters(@PathVariable UUID parameterUuid) {
```
antoinebhs, line 33: need userId + already exist in StudyController?

src/main/java/org/gridsuite/study/server/controller/sensitivityanalysis/SensitivityAnalysisStudyParametersController.java
```diff
@@ -0,0 +1,54 @@
...
+    private final SensitivityAnalysisService sensitivityAnalysisService;
+
+    public SensitivityAnalysisStudyParametersController(SensitivityAnalysisService sensitivityAnalysisService) {
+        this.sensitivityAnalysisService = sensitivityAnalysisService;
+    }
+
+    @GetMapping(value = "/parameters")
+    @Operation(summary = "Get sensitivity analysis parameters on study")
+    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "The sensitivity analysis parameters")})
+    public ResponseEntity<String> getSensitivityAnalysisParameters(
```
antoinebhs, line 39: sometimes it's annotated with produces json, sometimes not. make it consistent?

src/main/java/org/gridsuite/study/server/controller/dynamicmargincalculation/DynamicMarginCalculationParametersController.java
klesaulnier: Why do we need userId ? For now we don't check any authorization
The existing endpoint has been moved in src/main/java/org/gridsuite/study/server/controller/dynamicsimulation/DynamicSimulationStudyParametersController.java

But, it's not the same thing
- the old one is fetching parameter of a study
- the new one is fetching parameter directly from parameterUuid

src/main/java/org/gridsuite/study/server/controller/sensitivityanalysis/SensitivityAnalysisStudyParametersController.java
klesaulnier: It's already not consistant in existing code, but I'll match the new files endpoints behaviour and remove them from my new files

src/main/java/org/gridsuite/study/server/service/DynamicMappingService.java
```diff
@@ -59,4 +59,10 @@ public String getNetworkMatches(UUID networkUuid, String ruleToMatch) {
         HttpEntity<String> httpEntity = new HttpEntity<>(ruleToMatch, headers);
         return restTemplate.postForObject(url, httpEntity, String.class);
     }
+
+    public String getMappedModels(UUID mappingId) {
```
antoinebhs: is this used?

antoinebhs (approved):
We'll need to check if the errors are correctly forwarded
I think that later most endpoints will be migrated to *StudyParametersController* because we'll need the studyUuid to assert authz. Maybe we'll need to duplicate these endpoints in explore somehow? 
Some tests do not have much value but necessary for coverage...

src/main/java/org/gridsuite/study/server/service/DynamicMappingService.java
klesaulnier: removed

PR #1048, opened 2026-08-05
Poc gateway forward

klesaulnier:


antoinebhs (commented):
My opinion on this :
* it creates heterogenity in the server (with gateway/forwarded endpoints and "real endpoints"). I think there are existing endpoints that 
* it introduces some URI parsing that we wanted to avoid
* it needs some URI rewriting
One big advantage is that it avoid manual fowarding for all requests and eventual errors if forgetting request params etc

PR #1053, opened 2026-08-07
Use the modification applicability per root network tag

flomillot:
Depends on gridsuite/network-modification-server#867.

The applicabilities per root network are now held by the modifications themselves, keyed by root network tag, so study-server stops storing them:

- `RootNetworkNodeInfoEntity.modificationsUuidsToExclude` and everything that maintained it are removed. A modification now keeps its applicability wherever it goes.
- The application context and `BuildInfos` carry the tag of the root network instead of the excluded modification uuids.
- `GET /studies/{studyUuid}/nodes/{nodeUuid}/excluded-network-modifications` is removed and nothing replaces it: every modification carries its applicability per tag, so `GET /studies/{studyUuid}/nodes/{nodeUuid}/network-modifications` already relays it and the front end needs no second call.
- Changing the applicability of a shared modification requires the write permission on the shared element, since it is held by that element and used by every study referencing it.

- Renaming a root network tag, or deleting a root network, now reaches the applicabilities: they are keyed by the tag, so study-server calls the two new endpoints over every modification group of the study. The shared modification behind a reference is updated without checking the rights the user has on it, the rename only ever adding a tag there.

The applicabilities stored so far are not migrated: every modification goes back to applicable on every root network. Only production database will be migrated.


Mathieu-Deharbe (commented):
Not a bug but I think that this should be checked with the PO : 

If I rename the tag of a root network, the tags of the netmods remain the same and therefore the whole column is reactivated : 

Image

If this is the intented behavior feel free to resolve this thread.

src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -953,15 +953,15 @@ public ResponseEntity<String> getExportedNetworkModifications(@Parameter(descrip
         return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studyService.getExportedNetworkModifications(studyUuid, nodeUuid));
     }
 
-    @GetMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/excluded-network-modifications", produces = MediaType.APPLICATION_JSON_VALUE)
-    @Operation(summary = "Get excluded network modifications from a node")
-    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "The excluded network modifications were returned"), @ApiResponse(responseCode = "404",
+    @GetMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/network-modifications/applicability", produces = MediaType.APPLICATION_JSON_VALUE)
```
Mathieu-Deharbe, line 956: I think this endpoint is obsolete : why should we get the applicability separately and not with the network modifications themselves, in the same dto ? Now that all the data is in the same `ModificationInfos` class.

Mathieu-Deharbe (changes requested):

src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -1038,19 +1038,19 @@ public ResponseEntity<Void> updateNetworkModificationsMetadata(@Parameter(descri
...
-    public ResponseEntity<Void> updateNetworkModificationsActivation(@Parameter(description = "Study UUID") @PathVariable("studyUuid") UUID studyUuid,
+    public ResponseEntity<Void> updateNetworkModificationsApplicability(@Parameter(description = "Study UUID") @PathVariable("studyUuid") UUID studyUuid,
                                                                      @Parameter(description = "Root network UUID") @PathVariable("rootNetworkUuid") UUID rootNetworkUuid,
                                                                      @Parameter(description = "Node UUID") @PathVariable("nodeUuid") UUID nodeUuid,
                                                                      @Parameter(description = "Network modification UUIDs") @RequestParam("uuids") Set<UUID> networkModificationUuids,
-                                                                     @Parameter(description = "New activated value") @RequestParam(name = "activated") Boolean activated,
+                                                                     @Parameter(description = "New applicability value") @RequestParam(name = "applicable") Boolean applicable,
```
Mathieu-Deharbe: pplicable is the ability to be applied, which is true for all tags. I think for a boolean "applied" would be clearer.

src/main/java/org/gridsuite/study/server/controller/StudyController.java
flomillot: I didn't think about but you're right we can remove this useless API call.

src/main/java/org/gridsuite/study/server/service/NetworkModificationTreeService.java
```diff
@@ -186,7 +184,7 @@ private UUID duplicateNode(UUID nodeToCopyUuid, UUID anchorNodeUuid, InsertMode
         UUID newGroupUuid = UUID.randomUUID();
         UUID modificationGroupUuid = self.getModificationGroupUuid(nodeToCopyUuid);
         //First we create the modification group
-        Map<UUID, UUID> originToDuplicateModificationUuidMap = networkModificationService.duplicateModificationsGroup(modificationGroupUuid, newGroupUuid);
+        networkModificationService.duplicateModificationsGroup(modificationGroupUuid, newGroupUuid);
```
Mathieu-Deharbe, line 231: Now that the return value of `duplicateModificationsGroup` is not used anymore (here or anywhere else if I am right) I think it can be removed. 

Maybe even including from the endpoint in netmod-server but I wouldn't do it, just in case.

Mathieu-Deharbe (commented):
OK for me. I let the "request changes" tag because of https://github.com/gridsuite/study-server/pull/1053#pullrequestreview-4973746619

But for this we have to wait for users and PO input.

src/main/java/org/gridsuite/study/server/service/NetworkModificationTreeService.java
flomillot: Yes you're right, maybe we should also create a new pull request to update the endpoint in network modification server.

flomillot: https://github.com/gridsuite/network-modification-server/pull/875

src/main/java/org/gridsuite/study/server/service/NetworkModificationTreeService.java
```diff
@@ -821,17 +817,6 @@ public String getNetworkModifications(@NonNull UUID nodeUuid, boolean onlyStashe
         return networkModificationService.getModifications(self.getModificationGroupUuid(nodeUuid), onlyStashed, onlyMetadata);
     }
 
-    @Transactional
-    public List<ExcludedNetworkModifications> getModificationsToExclude(@NonNull UUID nodeUuid) {
-        List<RootNetworkNodeInfoEntity> rootNetworkByNodeInfos = rootNetworkNodeInfoService.getAllWithRootNetworkByNodeInfoId(nodeUuid);
```
Meklo, line 826: getAllWithRootNetworkByNodeInfoId has no longer any use, can be cleansed

src/main/java/org/gridsuite/study/server/service/NetworkModificationTreeService.java
```diff
@@ -1066,6 +1051,7 @@ private BuildInfos getBuildInfos(UUID nodeUuid, UUID rootNetworkUuid) {
                 throw new StudyException(BAD_NODE_TYPE, "The node " + entity.getIdNode() + " is not a modification node");
             } else {
                 buildInfos.setDestinationVariantId(self.getVariantId(nodeUuid, rootNetworkUuid));
+                buildInfos.setRootNetworkTag(rootNetworkService.getRootNetworkTag(rootNetworkUuid));
```
Meklo, line 1054: Why not placing it in insertModificationInfos as it was done for excluded modifications ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -369,10 +377,27 @@ public void createRootNetwork(@NonNull UUID studyUuid, @NonNull RootNetworkInfos
     }
 
     private void updateRootNetworkBasicInfos(UUID studyUuid, RootNetworkInfos rootNetworkInfos, boolean updateCase) {
+        String previousTag = rootNetworkService.getRootNetworkTag(rootNetworkInfos.getId());
         rootNetworkService.updateRootNetwork(rootNetworkInfos, updateCase);
```
Meklo, line 381: Shoudn't RootNetworkService::updateRootNetwork become transactional since you aim to read its new value immediatly after ?

src/main/java/org/gridsuite/study/server/service/NetworkModificationTreeService.java
flomillot: Ah yes I didn't see thanks.

flomillot: Ah yes I didn't see thanks.

src/main/java/org/gridsuite/study/server/service/NetworkModificationTreeService.java
flomillot: Because what we insert is not from modification info but it's the tag of the network.
To be honest I don't even know why it's called `insertModificationInfos`.

src/main/java/org/gridsuite/study/server/service/StudyService.java
flomillot: Actually, my call is already inside a transactional method, so with the persistence context, there is no issue. 
So my code doesn't introduce any issue, but the question is why it's not already a transactional.

Meklo: Surely an oversight

Meklo (approved):

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -313,8 +313,16 @@ public void deleteRootNetworks(UUID studyUuid, List<UUID> rootNetworksUuids, Str
...
 
+        List<String> deletedTags = allRootNetworkEntities.stream()
+                .filter(rootNetworkEntity -> rootNetworksUuids.contains(rootNetworkEntity.getId()))
+                .map(RootNetworkEntity::getTag)
+                .filter(Objects::nonNull)
+                .toList();
+
         rootNetworkService.deleteRootNetworks(studyEntity, rootNetworksUuids.stream());
 
+        networkModificationService.deleteRootNetworkTags(getStudyModificationGroupUuids(studyUuid), deletedTags);
```
Mathieu-Deharbe, line 324: Are you sure that we should delete tags from the netmods applicabilities when a root network is removed ?

**Edit : yes, the specs changed**

Mathieu-Deharbe (approved):

PR #1058, opened 2026-08-12
copy cut paste reference modification and add ref with target infos

souissimai:

### DIFF
--- a/src/main/java/org/gridsuite/study/server/dto/ReferenceData.java
+++ b/src/main/java/org/gridsuite/study/server/dto/ReferenceData.java
@@ -12,8 +12,8 @@ import java.util.UUID;
  * One occurrence of a network modification referencing a shared composite modification
  *
  * @param modificationUuid uuid of the modification-reference itself
- * @param referenceId uuid of the referenced shared composite
+ * @param sharedCompositeId uuid of the referenced shared composite
  * @param containerId uuid of the composite containing the reference, null if the modification-reference is at the root level
  */
-public record ReferenceData(UUID modificationUuid, UUID referenceId, UUID containerId) {
+public record ReferenceData(UUID modificationUuid, UUID sharedCompositeId, UUID containerId) {
 }
--- a/src/main/java/org/gridsuite/study/server/dto/ReferenceData.java
+++ b/src/main/java/org/gridsuite/study/server/dto/ReferenceData.java
@@ -12,8 +12,8 @@ import java.util.UUID;
  * One occurrence of a network modification referencing a shared composite modification
  *
  * @param modificationUuid uuid of the modification-reference itself
- * @param sharedCompositeId uuid of the referenced shared composite
+ * @param referenceId uuid of the referenced shared composite
  * @param containerId uuid of the composite containing the reference, null if the modification-reference is at the root level
  */
-public record ReferenceData(UUID modificationUuid, UUID sharedCompositeId, UUID containerId) {
+public record ReferenceData(UUID modificationUuid, UUID referenceId, UUID containerId) {
 }
--- a/src/main/java/org/gridsuite/study/server/dto/ReferenceData.java
+++ b/src/main/java/org/gridsuite/study/server/dto/ReferenceData.java
@@ -1,0 +1,20 @@
+/*
+  Copyright (c) 2026, RTE (http://www.rte-france.com)
+  This Source Code Form is subject to the terms of the Mozilla Public
+  License, v. 2.0. If a copy of the MPL was not distributed with this
+  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package org.gridsuite.study.server.dto;
+
+import java.util.UUID;
+
+/**
+ * One occurrence of a network modification referencing a shared composite modification stored in directory-server.
+ *
+ * @param modificationUuid uuid of the modification-reference itself
+ * @param referenceId uuid of the referenced shared composite in directory-server
+ * @param containerId uuid of the composite containing the reference, null if the modification-reference is at the root level
+ * @author RTE
+ */
+public record ReferenceData(UUID modificationUuid, UUID referenceId, UUID containerId) {
+}
--- a/src/main/java/org/gridsuite/study/server/dto/ReferenceData.java
+++ b/src/main/java/org/gridsuite/study/server/dto/ReferenceData.java
@@ -1,20 +1,19 @@
-/*
-  Copyright (c) 2026, RTE (http://www.rte-france.com)
-  This Source Code Form is subject to the terms of the Mozilla Public
-  License, v. 2.0. If a copy of the MPL was not distributed with this
-  file, You can obtain one at http://mozilla.org/MPL/2.0/.
+/**
+ * Copyright (c) 2026, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 package org.gridsuite.study.server.dto;
 
 import java.util.UUID;
 
 /**
- * One occurrence of a network modification referencing a shared composite modification stored in directory-server.
+ * One occurrence of a network modification referencing a shared composite modification
  *
  * @param modificationUuid uuid of the modification-reference itself
- * @param referenceId uuid of the referenced shared composite in directory-server
+ * @param referenceId uuid of the referenced shared composite
  * @param containerId uuid of the composite containing the reference, null if the modification-reference is at the root level
- * @author RTE
  */
 public record ReferenceData(UUID modificationUuid, UUID referenceId, UUID containerId) {
 }
--- a/src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveOrCopyInfos.java
+++ b/src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveOrCopyInfos.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright (c) 2026, RTE (http://www.rte-france.com)
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package org.gridsuite.study.server.dto.modification;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+import java.util.UUID;
+
+/**
+ * one entry of a copy/move-modifications request: pairs a modification with the container it currently belongs
+ * to, so a selection mixing modifications directly under a node's group and modifications nested in a composite
+ * can be resolved per-item, the same way {@link MoveModificationInfos} does for a single-modification move.
+ */
+public record ModificationMoveOrCopyInfos(
+        UUID modificationUuid,
+        @Schema(description = "container the modification currently belongs to; omitted only for GROUP, where node's own group is used") ModificationContainerInfos source) {
+}
--- a/src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveOrCopyInfos.java
+++ b/src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveOrCopyInfos.java
@@ -16,5 +16,5 @@ import java.util.UUID;
 
 public record ModificationMoveOrCopyInfos(
         UUID modificationUuid,
-        @Schema(description = "current source container of modification to copy/move") ModificationContainerInfos source) {
+        @Schema(description = "current source container of the moved/copied modification") ModificationContainerInfos source) {
 }
--- a/src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveOrCopyInfos.java
+++ b/src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveOrCopyInfos.java
@@ -11,7 +11,7 @@ import io.swagger.v3.oas.annotations.media.Schema;
 import java.util.UUID;
 
 /**
- * @author Maissa Souissi <abdelsalem.hedhili at rte-france.com>
+ * @author Maissa Souissi <maissa.souissi at rte-france.com>
  */
 
 public record ModificationMoveOrCopyInfos(
--- a/src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveOrCopyInfos.java
+++ b/src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveOrCopyInfos.java
@@ -1,4 +1,4 @@
-/*
+/**
  * Copyright (c) 2026, RTE (http://www.rte-france.com)
  * This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
@@ -11,11 +11,10 @@ import io.swagger.v3.oas.annotations.media.Schema;
 import java.util.UUID;
 
 /**
- * one entry of a copy/move-modifications request: pairs a modification with the container it currently belongs
- * to, so a selection mixing modifications directly under a node's group and modifications nested in a composite
- * can be resolved per-item, the same way {@link MoveModificationInfos} does for a single-modification move.
+ * @author Maissa Souissi <abdelsalem.hedhili at rte-france.com>
  */
+
 public record ModificationMoveOrCopyInfos(
         UUID modificationUuid,
-        @Schema(description = "container the modification currently belongs to; omitted only for GROUP, where node's own group is used") ModificationContainerInfos source) {
+        @Schema(description = "current source container of modification to copy/move") ModificationContainerInfos source) {
 }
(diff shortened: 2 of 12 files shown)


src/main/java/org/gridsuite/study/server/service/DirectoryService.java
```diff
@@ -185,4 +186,34 @@ public void checkPermission(List<UUID> elementUuids, UUID targetDirectoryUuid, S
...
+
+    /**
+     * moves node references of shared composite modifications in directory server: the existing references
+     * pointing to the origin node are updated to point to the target node (cut-paste: references move, they aren't duplicated)
+     * @param elementsUuids element uuids of the shared composites in directory server
+     * @param userId id of the user who moves the references
+     * @param originReferenceUuid the node references to update
+     * @param targetReferenceUuid where the references will point after the move
+     */
+    public void updateReferencesToSharedComposites(@NonNull List<UUID> elementsUuids, String userId, @NonNull UUID originReferenceUuid, UUID targetReferenceUuid) {
```
Mathieu-Deharbe: Seems to work fine but should include the `ReferenceType` of the targetReference.

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2117,6 +2124,19 @@ public void moveNetworkModifications(
...
+     * to the target node, instead of creating a new one (unlike the copy-paste flow).
+     */
+    private void updateReferencesToSharedComposites(Map<UUID, UUID> referenceTargets, String userId, UUID originNodeUuid, UUID targetNodeUuid) {
+        directoryService.updateReferencesToSharedComposites(
+                referenceTargets.keySet().stream().toList(),
+                userId,
+                originNodeUuid,
+                targetNodeUuid
+        );
+    }
```
Mathieu-Deharbe: This would only work when the target is a node. When it is a composite, the updating will need to change the referenceType which might be STUDY_NODE or become a NETWORK_MODIFICATION inside a composite. Same problem if the origin is a composite and the target is the root of a node.

referenceTargets contains its "mother" composite as value in the pair if there is one, null otherwise.

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2154,11 +2174,20 @@ public void duplicateNetworkModifications(
             (groupUuid, modificationApplicationContexts) -> {
                 NetworkModificationsResult networkModificationResults = networkModificationService.duplicateModifications(groupUuid, Pair.of(modificationsUuids, modificationApplicationContexts));
                 copyModificationsToExclude(originNodeUuid, targetNodeUuid, modificationsUuids, networkModificationResults);
+                createReferencesToSharedComposites(networkModificationService.getReferences(modificationsUuids), userId, targetNodeUuid);
```
Mathieu-Deharbe, line 2236: This seems to work but it probably won't when the duplicated `modificationsUuids` contain composites who contain reference modifications. Those will be duplicated too and references will have to be created for them as well, not just for `modificationsUuids`. 

So I think that we need a recursive version of `networkModificationService.getReferences`.

There is a TODO in netmod server about this : 
```
// TODO GRD-4785 : for now shared modification are only at the root level and can't be inside composites, so the composite uuid is set to null
        // but when it will be the case a specific function will have to be done in order to fetch the composite containing the modificationReference (if there is one)
```

Mathieu-Deharbe (changes requested):

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
souissimai: I added targetReferenceType

src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveOrCopyInfos.java
```diff
@@ -0,0 +1,20 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+package org.gridsuite.study.server.dto.modification;
+
+import io.swagger.v3.oas.annotations.media.Schema;
+
+import java.util.UUID;
+
+/**
+ * @author Maissa Souissi <abdelsalem.hedhili at rte-france.com>
```
Mathieu-Deharbe: Your mail adress is weird.

src/main/java/org/gridsuite/study/server/dto/ReferenceData.java
```diff
@@ -0,0 +1,19 @@
...
+import java.util.UUID;
+
+/**
+ * One occurrence of a network modification referencing a shared composite modification
+ *
+ * @param modificationUuid uuid of the modification-reference itself
+ * @param referenceId uuid of the referenced shared composite
+ * @param containerId uuid of the composite containing the reference, null if the modification-reference is at the root level
+ */
+public record ReferenceData(UUID modificationUuid, UUID referenceId, UUID containerId) {
```
Mathieu-Deharbe, line 18: I think this is clearer : 
```suggestion
 * @param sharedCompositeId uuid of the referenced shared composite
 * @param containerId uuid of the composite containing the reference, null if the modification-reference is at the root level
 */
public record ReferenceData(UUID modificationUuid, UUID sharedCompositeId, UUID containerId) {
```

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
```diff
@@ -185,4 +186,35 @@ public void checkPermission(List<UUID> elementUuids, UUID targetDirectoryUuid, S
 
         restTemplate.exchange(getDirectoryServerServerBaseUri() + path, HttpMethod.GET, new HttpEntity<>(headers), Void.class);
     }
+
+    /**
+     * update references of shared composite modifications in directory serve
```
Mathieu-Deharbe:
```suggestion
     * update references of shared composite modifications in directory server
```

src/main/java/org/gridsuite/study/server/service/DirectoryService.java
```diff
@@ -185,4 +186,35 @@ public void checkPermission(List<UUID> elementUuids, UUID targetDirectoryUuid, S
...
+     * @param elementsUuids element uuids of the shared composites
+     * @param userId id of the user who moves the references
+     * @param originReferenceUuid the node references to update
+     * @param targetReferenceUuid where the references will point after the move
```
Mathieu-Deharbe: Maybe ? : 
```suggestion
     * @param originReferenceUuid uuid of the container whose references are updated
     * @param targetReferenceUuid where the references will point after the update
```

src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveOrCopyInfos.java
```diff
@@ -0,0 +1,20 @@
...
+
+import java.util.UUID;
+
+/**
+ * @author Maissa Souissi <maissa.souissi at rte-france.com>
+ */
+
+public record ModificationMoveOrCopyInfos(
+        UUID modificationUuid,
+        @Schema(description = "current source container of modification to copy/move") ModificationContainerInfos source) {
```
Mathieu-Deharbe:
```suggestion
        @Schema(description = "current source container of the moved/copied modification") ModificationContainerInfos source) {
```
Otherwise the sentence may sound like the container is the one being moved.

src/main/java/org/gridsuite/study/server/service/RebuildNodeService.java
```diff
@@ -91,16 +90,10 @@ public UUID assembleModificationsIntoComposite(UUID studyUuid, UUID nodeUuid, Li
                 });
     }
 
-    private void handleMoveNetworkModifications(UUID studyUuid, UUID targetNodeUuid, UUID originNodeUuid, List<UUID> modificationsToCopyUuidList, String userId) {
+    private void handleMoveNetworkModifications(UUID studyUuid, UUID targetNodeUuid, UUID originNodeUuid, List<ModificationMoveOrCopyInfos> modificationInfos, String userId) {
```
Mathieu-Deharbe:
```suggestion
    private void handleMoveNetworkModifications(UUID studyUuid, UUID targetNodeUuid, UUID originNodeUuid, List<ModificationMoveOrCopyInfos> moveOrCopyInfos, String userId) {
```

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1798,55 +1801,151 @@ public RootNetworkIndexationStatus getRootNetworkIndexationStatus(UUID studyUuid
...
-            notificationService.emitStartModificationEquipmentNotification(studyUuid, originNodeUuid, originNodeChildrenUuids, NotificationService.MODIFICATIONS_UPDATING_IN_PROGRESS);
-        }
+        Map<ModificationContainerInfos, UUID> originNodeBySource = new LinkedHashMap<>();
+        modificationUuidsBySource.keySet().forEach(source ->
+                originNodeBySource.put(source, networkModificationTreeService.getNodeUuidByModificationGroup(source.id())));
+        Set<UUID> originNodesTouched = originNodeBySource.values().stream()
+                .filter(node -> node != null && !node.equals(targetNodeUuid))
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+
+        List<UUID> targetChildrenUuids = networkModificationTreeService.getChildrenUuids(targetNodeUuid);
```
Mathieu-Deharbe, line 1768: Just because I am getting lost in all the uuids : 
```suggestion
        List<UUID> targetChildrenNodeUuids = networkModificationTreeService.getChildrenUuids(targetNodeUuid);
```

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1798,55 +1801,151 @@ public RootNetworkIndexationStatus getRootNetworkIndexationStatus(UUID studyUuid
...
+        Map<UUID, List<UUID>> originChildrenUuidsByNode = originNodesTouched.stream()
+                .collect(Collectors.toMap(Function.identity(), networkModificationTreeService::getChildrenUuids, (a, b) -> a, LinkedHashMap::new));
+
+        notificationService.emitStartModificationEquipmentNotification(studyUuid, targetNodeUuid, targetChildrenUuids, NotificationService.MODIFICATIONS_UPDATING_IN_PROGRESS);
+        originChildrenUuidsByNode.forEach((originNodeUuid, children) ->
+                notificationService.emitStartModificationEquipmentNotification(studyUuid, originNodeUuid, children, NotificationService.MODIFICATIONS_UPDATING_IN_PROGRESS));
```
Mathieu-Deharbe: That part about notifications could probably be extracted for clarity, but we will see in the refacto.

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1798,55 +1801,151 @@ public RootNetworkIndexationStatus getRootNetworkIndexationStatus(UUID studyUuid
...
+                                                    UUID targetNodeUuid, ReferenceAttributes.ReferenceType targetReferenceType) {
+        List<UUID> referenceUuids = referenceTargets.stream()
+                .map(ReferenceData::referenceId)
+                .collect(Collectors.toList());
+        directoryService.updateReferencesToSharedComposites(
+                referenceUuids,
+                userId,
+                originNodeUuid,
+                targetNodeUuid,
+                targetReferenceType
```
Mathieu-Deharbe: Those may be composite uuids, not node uuids : 
```suggestion
    private void updateReferenceWhenMoveModification(List<ReferenceData> referenceTargets, String userId, UUID originReferenceUuid,
                                                    UUID targetReferenceUuid, ReferenceAttributes.ReferenceType targetReferenceType) {
        List<UUID> referenceModificationsUuids = referenceTargets.stream()
                .map(ReferenceData::referenceId)
                .collect(Collectors.toList());
        directoryService.updateReferencesToSharedComposites(
                referenceModificationsUuids,
                userId,
                originReferenceUuid,
                targetReferenceUuid,
                targetReferenceType
```

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1798,55 +1801,151 @@ public RootNetworkIndexationStatus getRootNetworkIndexationStatus(UUID studyUuid
...
+                : networkModificationService.findParentComposites(modificationUuidsNeedingSourceLookup);
+
+        return modificationInfos.stream()
+                .collect(Collectors.groupingBy(
+                        info -> resolveSourceContainer(info.source(), info.modificationUuid(), fallbackSourceNodeUuid, parentCompositeByModificationUuid),
+                        LinkedHashMap::new,
+                        Collectors.mapping(ModificationMoveOrCopyInfos::modificationUuid, Collectors.toList())));
+    }
+
+    private boolean hasModificationSource(ModificationContainerInfos source) {
```
Mathieu-Deharbe, line 1879: Given that the cource may not be a modification and no data here involves a modification : 
```suggestion
    private boolean hasSource(ModificationContainerInfos source) {
```
or
```suggestion
    private boolean sourceDefined(ModificationContainerInfos source) {
```

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1870,18 +1969,81 @@ private void emitNetworkModificationImpactsForAllRootNetworks(List<Optional<Netw
...
+                    return networkModificationResults;
+                },
+                userId);
+    }
+
+    /**
+     * @return old modification uuid -> new (copied) modification uuid, for every modification duplicated by
+     * {@code networkModificationResults} - root-level ones as well as those nested in a duplicated composite
+     */
+    private Map<UUID, UUID> buildModificationsUuidMapping(List<UUID> modificationsUuids, List<UUID> originalChildrenUuids, NetworkModificationsResult networkModificationResults) {
```
Mathieu-Deharbe: This should make the correspondence between this and the copyUuids clearer : 
```suggestion
    private Map<UUID, UUID> buildModificationsUuidMapping(List<UUID> copiedUuids, List<UUID> copiedChildrenUuids, NetworkModificationsResult networkModificationResults) {
```

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1870,18 +1969,81 @@ private void emitNetworkModificationImpactsForAllRootNetworks(List<Optional<Netw
...
+
+    /**
+     * one new reference is created per pasted modification-reference occurrence: even if several of them target the
+     * same shared composite, each occurrence must get its own reference row in directory-server (symmetric with
+     * {@link #removeReferences}, which likewise issues one removal per occurrence on delete) - so this must NOT be
+     * deduplicated by referenceId.
+     */
+    private void createReferencesToSharedComposites(List<ReferenceData> referenceTargets, List<UUID> modificationsUuids,
+                                                    Map<UUID, UUID> mappingModificationsUuids, String userId, UUID targetNodeUuid) {
+        Set<UUID> requestedUuids = new HashSet<>(modificationsUuids);
```
Mathieu-Deharbe: I don't think these are targets : 
```suggestion
    private void createReferencesToSharedComposites(List<ReferenceData> references, List<UUID> modificationsUuids,
                                                    Map<UUID, UUID> mappingModificationsUuids, String userId, UUID targetNodeUuid) {
        Set<UUID> requestedUuids = new HashSet<>(modificationsUuids);
```
I understand the function but this is very convoluted and looks like the data could be stored in `mappingModificationsUuids` when `buildModificationsUuidMapping` is called. With a specific mapping structure we could probably simplify these functions and comments. modificationsUuids would be removed etc...

I am not asking you to do this here but could you reference this in the refacto ticket ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1870,18 +1969,81 @@ private void emitNetworkModificationImpactsForAllRootNetworks(List<Optional<Netw
...
+
+        List<UUID> copyChildren = networkModificationService.findAllChildrenUuids(copyUuids);
+        for (int i = 0; i < originalChildrenUuids.size(); i++) {
+            mappingModificationsUuids.put(originalChildrenUuids.get(i), copyChildren.get(i));
+        }
+
+        return mappingModificationsUuids;
+    }
+
+    private void copyModificationsToExclude(UUID originNodeUuid, UUID targetNodeUuid, Map<UUID, UUID> mappingModificationsUuids) {
```
Mathieu-Deharbe, line 1962: You should remove the original copyModificationsToExclude function now useless; Anyway this will all be removed by Florent ticket about applicability.

src/test/java/org/gridsuite/study/server/RebuildNodeServiceTest.java
```diff
@@ -42,6 +43,11 @@ class RebuildNodeServiceTest {
     @MockitoBean
     private StudyService studyService;
 
+    // RebuildNodeService's bulk move path now looks up each modification's real parent composite
+    // before grouping by source; mock it here too so these tests don't hit the real HTTP client
```
Mathieu-Deharbe: Remove the comment or put this before : `doReturn(Map.of()).when(networkModificationService).findParentComposites(any());`

Mathieu-Deharbe (commented):
Only small remarks but I didn't validate yet because I need to test one last time after the conflict solving.

src/main/java/org/gridsuite/study/server/service/StudyService.java
souissimai: done

src/main/java/org/gridsuite/study/server/service/StudyService.java
souissimai: ok

src/main/java/org/gridsuite/study/server/service/StudyService.java
souissimai: here we create refernce for list of copied modification , gourped by source container, exact !to refactor ! created ticket

Mathieu-Deharbe (approved):

PR #1067, opened 2026-08-19
network load status

ghazwarhili:
## PR Summary
network load status

### DIFF
--- a/src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
+++ b/src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
@@ -188,4 +188,11 @@ public class SupervisionController {
         return ResponseEntity.ok().build();
     }
 
+    @GetMapping(value = "/studies/loaded")
+    @Operation(summary = "Get the study uuids whose network is currently loaded")
+    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "List of the study uuids whose network is currently loaded")})
+    public ResponseEntity<List<UUID>> getLoadedStudies(@Parameter(description = "Study uuids to filter") @RequestParam("ids") List<UUID> studyUuids) {
+        return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(supervisionService.getLoadedStudyUuids(studyUuids));
+    }
+
 }
--- a/src/main/java/org/gridsuite/study/server/service/StudyService.java
+++ b/src/main/java/org/gridsuite/study/server/service/StudyService.java
@@ -19,6 +19,7 @@ import org.gridsuite.study.server.StudyConstants;
 import org.gridsuite.study.server.dto.*;
 import org.gridsuite.study.server.dto.InvalidateNodeTreeParameters.ComputationsInvalidationMode;
 import org.gridsuite.study.server.dto.InvalidateNodeTreeParameters.InvalidationMode;
+import org.gridsuite.study.server.dto.NetworkLoadStatus;
 import org.gridsuite.study.server.dto.caseimport.CaseImportAction;
 import org.gridsuite.study.server.dto.computation.ComputationParameterUUIDs;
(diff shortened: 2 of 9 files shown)


src/main/java/org/gridsuite/study/server/service/SupervisionService.java
```diff
@@ -387,10 +382,16 @@ public void invalidateStudy(UUID studyUuid) {
             var rootNodeUuid = networkModificationTreeService.getStudyRootNodeUuid(studyUuid);
             studyService.unblockNodeTree(studyUuid, rootNodeUuid);
         }
-        notificationService.emitElementUpdated(studyUuid, SUPERVISION_USER);
         LOGGER.trace("Study {} nodes builds deleted and root node invalidated in : {} milliseconds", studyUuid, TimeUnit.NANOSECONDS.toMillis(System.nanoTime() - startTime.get()));
     }
 
+    @Transactional(readOnly = true)
+    public List<UUID> getLoadedStudyUuids(List<UUID> studyUuids) {
```
Meklo, line 416: Better fitted in StudyService no ? It may have other usages outside of supervision scope

src/main/java/org/gridsuite/study/server/controller/SupervisionController.java
```diff
@@ -188,4 +188,11 @@ public ResponseEntity<Void> invalidateStudy(@PathVariable("studyUuid") UUID stud
         return ResponseEntity.ok().build();
     }
 
+    @GetMapping(value = "/studies/loaded")
+    @Operation(summary = "Get the study uuids whose network is currently loaded")
+    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "List of the study uuids whose network is currently loaded")})
+    public ResponseEntity<List<UUID>> getLoadedStudies(@Parameter(description = "Study uuids to filter") @RequestParam("ids") List<UUID> studyUuids) {
```
Meklo, line 194: Could be moved to study controller as well no ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2993,12 +3001,14 @@ public Map<ComputationType, String> getAllComputationsStatus(@NonNull UUID study
...
         rootNetworkService.assertIsRootNetworkInStudy(studyUuid, rootNetworkUuid);
+        self.updateNetworkLoadStatus(studyUuid, NetworkLoadStatus.UNLOADING);
         var rootNodeUuid = networkModificationTreeService.getStudyRootNodeUuid(studyUuid);
         // First we unbuild all nodes
         doUnbuildNodeTree(studyUuid, rootNodeUuid, true, true, userId);
         // Then we erase data linked to root node on all root networks
         rootNetworkService.invalidateRootNetworkRemoteInfos(List.of(rootNetworkService.getRootNetworkInfos(rootNetworkUuid)), true, false);
         rootNetworkService.updateRootNetworkIndexationStatus(studyUuid, rootNetworkUuid, RootNetworkIndexationStatus.NOT_INDEXED);
+        self.updateNetworkLoadStatus(studyUuid, NetworkLoadStatus.UNLOADED);
         notificationService.emitRootNetworksUpdated(studyUuid);
```
Meklo, line 2726: Those load status update might be better placed in one level above in invalidateStudy. Otherwise the study will switch to UNLOADING multiple times during an invalidation with multiple root networks. Also, in a multi root network situation, a study might have the UNLOADED status even if some root networks failed to unload

Meklo (commented):
It might be worth to also add or tweak a reindexation test to check if the study goes through UNLOADED -> LOADED status

src/main/java/org/gridsuite/study/server/service/StudyService.java
Meklo: Plus it would avoid using self as a workaround for transaction proxy

PR #1070, opened 2026-08-19
Add an endpoint sharing a composite modification of a node

flomillot:
Adds `POST /v1/studies/{studyUuid}/nodes/{nodeUuid}/network-modifications/{modificationUuid}/share?name=&description=&parentDirectoryUuid=`, which moves a composite modification of a node out of the study into a directory of GridExplore, and replaces it in the node by a reference to this now shared modification.

The write permission on the target directory and the uniqueness of the name are checked before anything is moved. The composite is then extracted by the network modification server — keeping its uuid — stored as a `MODIFICATION` element of the directory server, and referenced from the node.

Since the applied modifications are left unchanged, the node is not invalidated: only the modifications update notifications are emitted, so the front refetches the list.

Requires gridsuite/network-modification-server#874.


GhilesA (approved):
Just apply the sonar recommandation on QUERY_PARAM_GROUP_UUID

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2180,6 +2180,40 @@ public UUID assembleModificationsIntoComposite(
...
+            // the composite keeps its uuid when extracted, the applied modifications are left unchanged : the node
+            // does not need to be rebuilt
+            UUID sharedCompositeUuid = networkModificationService.extractCompositeModificationToShare(groupUuid, modificationUuid, name);
+            directoryService.createElement(parentDirectoryUuid, description, sharedCompositeUuid, name, DirectoryService.MODIFICATION, userId);
+            directoryService.createsReferencesToSharedComposites(List.of(sharedCompositeUuid), userId, nodeUuid);
+        } finally {
+            notificationService.emitEndModificationEquipmentNotification(studyUuid, nodeUuid, childrenUuids);
+        }
+        notificationService.emitElementUpdated(studyUuid, userId);
+    }
```
ghazwarhili, line 2215: missing catch in case of failures midway ! 
we have to handle the rollback in case of error

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2180,6 +2180,40 @@ public UUID assembleModificationsIntoComposite(
         return newCompositeUuid;
     }
 
+    /**
+     * Moves a composite modification of a node out of the study : it is stored as an element in the directory server,
+     * and replaced in the node by a reference to this now shared composite modification.
+     */
+    @Transactional
```
ghazwarhili: why @Transactional here ?
i can't see any JPA transaction here

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2180,6 +2180,40 @@ public UUID assembleModificationsIntoComposite(
...
+        UUID parentDirectoryUuid,
+        String userId) {
+        // checks we can write in the target directory before touching anything
+        directoryService.checkPermission(List.of(), parentDirectoryUuid, userId, PermissionType.WRITE, false);
+        if (directoryService.elementExists(parentDirectoryUuid, name, DirectoryService.MODIFICATION)) {
+            throw new StudyException(ELEMENT_ALREADY_EXISTS, "composite modification name " + name + " already exists in directory", Map.of("fileName", name));
+        }
+
+        UUID groupUuid = networkModificationTreeService.getModificationGroupUuid(nodeUuid);
+        List<UUID> childrenUuids = networkModificationTreeService.getChildrenUuids(nodeUuid);
```
ghazwarhili, line 2203: maybe you need to add @Transactional(readOnly = true)

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -2180,6 +2180,40 @@ public UUID assembleModificationsIntoComposite(
...
+        notificationService.emitStartModificationEquipmentNotification(studyUuid, nodeUuid, childrenUuids, NotificationService.MODIFICATIONS_UPDATING_IN_PROGRESS);
+        try {
+            // the composite keeps its uuid when extracted, the applied modifications are left unchanged : the node
+            // does not need to be rebuilt
+            UUID sharedCompositeUuid = networkModificationService.extractCompositeModificationToShare(groupUuid, modificationUuid, name);
+            directoryService.createElement(parentDirectoryUuid, description, sharedCompositeUuid, name, DirectoryService.MODIFICATION, userId);
+            directoryService.createsReferencesToSharedComposites(List.of(sharedCompositeUuid), userId, nodeUuid);
```
ghazwarhili: if directory server fails to create element and the fails to createsReferencesToSharedComposites
=> orphaned directory element !

ghazwarhili (changes requested):

src/main/java/org/gridsuite/study/server/service/StudyService.java
flomillot: There is some reading. I will add read only.

src/main/java/org/gridsuite/study/server/service/StudyService.java
flomillot: Yes, that's what I did, good catch.

src/main/java/org/gridsuite/study/server/service/StudyService.java
flomillot: In theory I agree with you but doesn't seem like it's what we do usually in the application. Can you show me where it's done actually in the service, as an exemple ? I mostly see logs.

src/main/java/org/gridsuite/study/server/service/StudyService.java
flomillot: Yes, and it's already the case everywhere in Study serverr, see `insertCompositeNetworkModifications` for instance., which is pretty close.
The rollback is actually not managed in Study server unlike in Explore server (cf executeWithRollback).
I think it need a dedicated technical story.

ghazwarhili: No dedicated technical story needed here IMO so wrapping createsReferencesToSharedComposites + extractCompositeModificationToShare in a try/catch right after createElement with directoryService.deleteElement in the catch, is enough

src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
```diff
@@ -410,6 +412,26 @@ public NetworkModificationsResult insertCompositeModifications(UUID groupUuid,
         ).getBody();
     }
 
+    /**
+     * Asks the network modification server to take a composite modification out of its group, replacing it in the group
+     * by a reference to it, so that it can be stored as an element in the directory server.
+     * @return the uuid of the extracted composite modification
+     */
+    public UUID extractCompositeModificationToShare(@NonNull UUID groupUuid, @NonNull UUID modificationUuid, @NonNull String name) {
```
ghazwarhili: why return UUID ?
the implementation of extractCompositeModificationToShare in NetworkModificationRepository return uuid which is always equal to modificationUuid (the composite entity keeps its own id only its container changes )

src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -687,6 +687,23 @@ public ResponseEntity<UUID> assembleModificationsIntoNewComposite(
         return ResponseEntity.ok().body(newCompositeUuid);
     }
 
+    @PostMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/network-modifications/{modificationUuid}/share")
+    @Operation(summary = "Move a composite modification of the node into a directory, replacing it by a reference to it")
```
ghazwarhili, line 691: it is about moving or extract or share ?

@Operation(summary = "Extract the composite modification and replace it with a reference") it is better ? what do you think

src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
flomillot: It's an API good practice, when you use a POST you always return the ID or the created element.
I hesitated to return the reference modification ID but actually what we create and the most important is the shared modification.

src/main/java/org/gridsuite/study/server/controller/StudyController.java
flomillot: Well when you share a modification you extract it from the node and move it into the directory so all three.
I prefer move for API description however.

src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
flomillot: But indeed we don't use this a lot, so maybe I should not return anything.

src/main/java/org/gridsuite/study/server/service/StudyService.java
flomillot: I'll ask a TL about it, because if we do it we should do something like in explore server which is not a quick fix.

ghazwarhili: @etiennehomer

flomillot: We let like this for now

ghazwarhili (approved):

PR #1077, opened 2026-09-01
Razwa/fix duplicate study root network not found

ghazwarhili:
## PR Summary
duplicateStudyRootNetworks previously always cloned the source network in network store, even when the source root network's network had already been unloaded (RootNetworkLoadStatus.UNLOADED`). Since the underlying network no longer exists in that case, cloneNetwork/getNetworkVariants failed with a root network not found error                                                                  
                                                                                                                                                                                                                                                                                                                                                                                                           
  - Check networkService.doesNetworkExist(sourceNetworkUuid) before cloning if the network doesn't exist skip variant lookup/cloning and reuse the source network UUID as is marking the new root network UNLOADED too                                                                                                                                                                                          
  - reindexRootNetwork now skips reindexing (and marks indexation NOT_INDEXED) when the root network isn't LOADED instead of failing on a missing network


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1707,7 +1707,11 @@ private void reindexRootNetwork(StudyEntity study, UUID rootNetworkUuid) {
         // reindex root network for study in elasticsearch
         studyInfosService.recreateStudyInfos(studyInfos);
         RootNetworkEntity rootNetwork = rootNetworkService.getRootNetwork(rootNetworkUuid).orElseThrow(() -> new StudyException(NOT_FOUND, "Root network not found"));
-
+        if (rootNetwork.getLoadStatus() != RootNetworkLoadStatus.LOADED) {
+            updateRootNetworkIndexationStatus(study, rootNetwork, RootNetworkIndexationStatus.NOT_INDEXED);
```
etiennehomer: to remove

ghazwarhili: removed

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1707,7 +1707,10 @@ private void reindexRootNetwork(StudyEntity study, UUID rootNetworkUuid) {
         // reindex root network for study in elasticsearch
         studyInfosService.recreateStudyInfos(studyInfos);
         RootNetworkEntity rootNetwork = rootNetworkService.getRootNetwork(rootNetworkUuid).orElseThrow(() -> new StudyException(NOT_FOUND, "Root network not found"));
-
+        if (rootNetwork.getLoadStatus() != RootNetworkLoadStatus.LOADED) {
+            LOGGER.info("Root network '{}' has no loaded network, skipping reindexation", rootNetworkUuid);
```
etiennehomer:
```suggestion
            LOGGER.info("Root network '{}' is not loaded, skipping reindexation", rootNetworkUuid);
```

etiennehomer (changes requested):

src/main/java/org/gridsuite/study/server/service/StudyService.java
ghazwarhili: done

etiennehomer (approved):

PR #1081, opened 2026-09-04
Fix race condition by synchronizing StudyServerExecutionService for u…

SlimaneAmar:
…nit tests


sBouzols (approved):
Code review OK
Needed since `invalidateNodeTree` calls `deleteInvalidationInfos` which use the `StudyServerExecutionService` 👍

PR #1082, opened 2026-09-04
Change checkNetworkExistence endpoint's body to include network load status

basseche:
## PR Summary

In order to differentiate, when network does not exist in store, between loaded and unloaded network
we change the endpoint checkNetworkExistence., to add the information on root network load status.


src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -279,17 +279,15 @@ public ResponseEntity<Void> cutAndPasteNode(@PathVariable("studyUuid") UUID stud
...
         @ApiResponse(responseCode = "204", description = "The network doesn't exist")})
-    public ResponseEntity<Void> checkNetworkExistence(@PathVariable("studyUuid") UUID studyUuid, @PathVariable("rootNetworkUuid") UUID rootNetworkUuid) {
+    public ResponseEntity<RootNetworkExistence> checkNetworkExistence(@PathVariable("studyUuid") UUID studyUuid, @PathVariable("rootNetworkUuid") UUID rootNetworkUuid) {
         UUID networkUUID = rootNetworkService.getNetworkUuid(rootNetworkUuid);
-        return networkStoreService.doesNetworkExist(networkUUID)
-            ? ResponseEntity.ok().build()
-            : ResponseEntity.noContent().build();
-
+        return ResponseEntity.ok().body(new RootNetworkExistence(networkStoreService.doesNetworkExist(networkUUID),
+                rootNetworkService.getRootNetworkLoadStatus(rootNetworkUuid)));
```
ghazwarhili, line 290: you don't need to check whether the network exists on the store side? From my understanding, networkStoreService.doesNetworkExist(networkUUID) is not necessary and can be removed.
@etiennehomer what do you think ?

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
```diff
@@ -89,6 +89,11 @@ public UUID getNetworkUuid(UUID rootNetworkUuid) {
         return getRootNetwork(rootNetworkUuid).map(RootNetworkEntity::getNetworkUuid).orElseThrow(() -> new StudyException(NOT_FOUND, "Root network not found"));
     }
 
+    public RootNetworkLoadStatus getRootNetworkLoadStatus(UUID networkUuid) {
+        Optional<RootNetworkEntity> rootNetworkEntity = getRootNetwork(networkUuid);
+        return rootNetworkEntity.map(RootNetworkEntity::getLoadStatus).orElse(null);
```
ghazwarhili: why Optional here (else null) ?  (the default value is LOADED)

src/main/java/org/gridsuite/study/server/dto/RootNetworkExistence.java
```diff
@@ -0,0 +1,4 @@
+package org.gridsuite.study.server.dto;
+
+public record RootNetworkExistence(Boolean exists, RootNetworkLoadStatus rootNetworkLoadStatus) {
```
ghazwarhili: I'm not sure we need this DTO

src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -279,17 +279,15 @@ public ResponseEntity<Void> cutAndPasteNode(@PathVariable("studyUuid") UUID stud
...
     }
 
-    @RequestMapping(value = "/studies/{studyUuid}/root-networks/{rootNetworkUuid}/network", method = RequestMethod.HEAD)
+    @GetMapping(value = "/studies/{studyUuid}/root-networks/{rootNetworkUuid}/network")
     @Operation(summary = "check study root network existence")
     @ApiResponses(value = {
         @ApiResponse(responseCode = "200", description = "The network does exist"),
         @ApiResponse(responseCode = "204", description = "The network doesn't exist")})
-    public ResponseEntity<Void> checkNetworkExistence(@PathVariable("studyUuid") UUID studyUuid, @PathVariable("rootNetworkUuid") UUID rootNetworkUuid) {
+    public ResponseEntity<RootNetworkExistence> checkNetworkExistence(@PathVariable("studyUuid") UUID studyUuid, @PathVariable("rootNetworkUuid") UUID rootNetworkUuid) {
```
ghazwarhili: Would it be possible to return a simple boolean (true or false) instead?

src/main/java/org/gridsuite/study/server/dto/RootNetworkExistence.java
basseche: What do you mean ? You want to send a pair ? We can do that but it will be less clear.

src/main/java/org/gridsuite/study/server/controller/StudyController.java
basseche: No boolean would be for exists or not but What about the load status ? We need it in front side

src/main/java/org/gridsuite/study/server/service/RootNetworkService.java
basseche: but if networkEntity does not exist are you sure we should return Loaded ?

src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -279,17 +279,15 @@ public ResponseEntity<Void> cutAndPasteNode(@PathVariable("studyUuid") UUID stud
         return ResponseEntity.ok().build();
     }
 
-    @RequestMapping(value = "/studies/{studyUuid}/root-networks/{rootNetworkUuid}/network", method = RequestMethod.HEAD)
+    @GetMapping(value = "/studies/{studyUuid}/root-networks/{rootNetworkUuid}/network")
     @Operation(summary = "check study root network existence")
     @ApiResponses(value = {
         @ApiResponse(responseCode = "200", description = "The network does exist"),
         @ApiResponse(responseCode = "204", description = "The network doesn't exist")})
```
etiennehomer: to remove

src/main/java/org/gridsuite/study/server/dto/RootNetworkStatusInfos.java
```diff
@@ -0,0 +1,4 @@
+package org.gridsuite.study.server.dto;
+
```
ghazwarhili, line 8: author and licence

ghazwarhili (approved):

PR #1084, opened 2026-09-07
Ignore the applicabilities in a mono root network study

flomillot:
A study holding a single root network now resolves no tag at all: every activated modification applies, whatever the applicabilities it carries. Those are left untouched, hidden until the study turns multi root network.

Creating a second root network then invalidates what was built on the first one, since the applicabilities are taken into account from that point on.

### DIFF
--- a/src/main/java/org/gridsuite/study/server/service/StudyService.java
+++ b/src/main/java/org/gridsuite/study/server/service/StudyService.java
@@ -367,13 +367,23 @@ public class StudyService {
             rootNetworkService.createRootNetwork(studyEntity, rootNetworkInfos);
             rootNetworkService.deleteRootNetworkRequest(rootNetworkCreationRequestEntityOpt.get());
             //update study entity to multi root
-            if (studyEntity.getRootNetworks().size() > 1) {
+            if (studyEntity.isMonoRoot()) {
                 studyEntity.setMonoRoot(false);
+                invalidatePreviousRootNetworkNodeTree(studyEntity);
             }
         } else {
             rootNetworkService.deleteRootNetworks(studyEntity, List.of(rootNetworkInfos));
         }
         notificationService.emitRootNetworksUpdated(studyUuid);
+    }
+
+    /**
+     * The nodes built while the study was mono root network ignored the applicabilities, so they have to be built
+     * again now that these are taken into account.
+     */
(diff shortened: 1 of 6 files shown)


src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -367,15 +367,25 @@ public void createRootNetwork(@NonNull UUID studyUuid, @NonNull RootNetworkInfos
...
         notificationService.emitRootNetworksUpdated(studyUuid);
     }
 
+    /**
+     * The nodes built while the study was mono root network ignored the applicabilities, so they have to be built
+     * again now that these are taken into account.
+     */
+    private void invalidatePreviousRootNetworkNodeTree(StudyEntity studyEntity) {
+        UUID rootNodeUuid = networkModificationTreeService.getStudyRootNodeUuid(studyEntity.getId());
+        invalidateNodeTree(studyEntity.getId(), rootNodeUuid, studyEntity.getRootNetworks().getFirst().getId());
```
Meklo: We sure `studyEntity.getRootNetworks().getFirst().getId()` will always be consistent with the original root network ? 
Maybe moving the monoroot check and invalidatePreviousRootNetworkNodeTree invocation before rootNetworkService::createRootNetwork could prevent potential inconsistencies

Meklo (commented):
All good functionnally, maybe just a guard to be added as per comment 
I think there also was mention of adding a build invalidation warning in the root network creation dialog specified in the issue comments. It might also need a gridstudy PR

src/main/java/org/gridsuite/study/server/service/StudyService.java
flomillot: Yes right now it is.
And we don't want to switch to multi repo and invalidate the tree if the creation failed.
So I suggest to stay like that (I switched to getFirstRootNetwork, seems better).

flomillot:
> All good functionnally, maybe just a guard to be added as per comment I think there also was mention of adding a build invalidation warning in the root network creation dialog specified in the issue comments. It might also need a gridstudy PR

Yes I saw it after it was in discussion, here the PR https://github.com/gridsuite/gridstudy-app/pull/4190

Meklo (approved):

PR #1093, opened 2026-09-11
extract composite to share description

EtienneLt:

### DIFF
--- a/src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
+++ b/src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
@@ -52,6 +52,7 @@ public class NetworkModificationService {
     private static final String NETWORK_MODIFICATIONS_COUNT_PATH = "network-modifications-count";
     private static final String QUERY_PARAM_ACTION = "action";
     private static final String QUERY_PARAM_NAME = "name";
+    private static final String QUERY_PARAM_DESCRIPTION = "description";
     private static final String QUERY_PARAM_GROUP_UUID = "groupUuid";
     private static final String QUERY_PARAM_ROOT_NETWORK_TAG = "rootNetworkTag";
     private static final String QUERY_PARAM_GROUP_UUIDS = "groupUuids";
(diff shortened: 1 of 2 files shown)


src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
Mathieu-Deharbe: userId should be in the headers : 
```suggestion
            .buildAndExpand()
            .toUriString();

        HttpHeaders headers = new HttpHeaders();
        headers.set(HEADER_USER_ID, userId);
        headers.setContentType(MediaType.APPLICATION_JSON);
```

Mathieu-Deharbe (changes requested):
**From gridexplore** if I modify the description with many special characters (I added the description "snif&=??*&+*.*&?fds"), the description is not correctly updated **in gridstudy** : probably an encoding problem because the resulting description in gridstudy is only "snif" like if "&=??*&+*.*&?fds" was considered a secondary parameter.

src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
```diff
@@ -528,9 +530,10 @@ public NetworkModificationsResult insertCompositeModifications(UUID groupUuid,
      * @return the reference modification left in place of the composite modification, either in the group of the node
      * or in a parent composite
      */
-    public ModificationReference extractCompositeModificationToShare(@NonNull UUID groupUuid, @NonNull UUID modificationUuid, @NonNull String name) {
+    public ModificationReference extractCompositeModificationToShare(@NonNull UUID groupUuid, @NonNull UUID modificationUuid, @NonNull String name, String description) {
         String path = UriComponentsBuilder.fromPath(COMPOSITE_PATH + "{modificationUuid}" + DELIMITER + "share")
                 .queryParam(QUERY_PARAM_NAME, name)
+                .queryParam(QUERY_PARAM_DESCRIPTION, description)
```
Mathieu-Deharbe: The description parameter problem probably happens here. I guess it should be encoded or put in the body.

PR #1098, opened 2026-09-15
Handle move in batch and dissociate from copy endpoint

Meklo:


src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -651,46 +653,57 @@ public ResponseEntity<String> getAllMapData(
...
+                        studyService.assertIsNodeExist(studyUuid, id);
+                    } catch (StudyException _) {
+                        throw new StudyException(MOVE_NETWORK_MODIFICATION_FORBIDDEN);
+                    }
+                });
+
+        // Origin node for rebuild: first foreign source node, or same node
+        UUID originNodeUuid = modificationMoveRequests.stream()
+                .map(ModificationMoveRequest::source)
+                .map(ModificationLocationInfos::nodeUuidOrNull)
```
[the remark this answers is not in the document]
Meklo: This data isn't held in study-server and it would feel cluttered to add a remote call just for this check. Would always passing the node uuid along the composite in the location dto solve the issue ?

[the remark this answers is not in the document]
Meklo: We trust the data provided by the app but I agree it's an issue worth fixing in the future. I've systematically added the nodeUuid along the composites uuid so their node will be validated against the study 

Moreso the rebuilding process isn't able to handle multiple origin nodes which the app won't do atm also.
Is it worth staying in this state while adding this comment ? :
```
        // TODO The system isn't currently able to properly handle rebuild with multiple different node sources
        // which isn't a feature in the app atm, need some adaptation if multiple source nodes need rebuilding
        // thus the origin node for rebuild is the first foreign source node, or the node at hand
```

src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveInfos.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.UUID;
+
+/**
+ * @author Hugo Marcellin <hugo.marcelin at rte-france.com>
+ */
+public record ModificationMoveInfos(
+        UUID modificationUuid,
+        @Schema(description = "current container; resolved to the parent composite or the origin node's group when omitted") ModificationContainerInfos source,
+        @Schema(description = "destination container; defaults to the target node's group") ModificationContainerInfos target,
+        @Schema(description = "insert before this modification of the target container; appends when null") UUID beforeUuid) { }
```
basseche: here it is less ambiguous since there is a description, you should maybe rename it also if you rename the one in network-modification-server

basseche (approved):

src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -651,46 +653,59 @@ public ResponseEntity<String> getAllMapData(
         return ResponseEntity.ok().contentType(MediaType.APPLICATION_JSON).body(studyService.getAllMapData(studyUuid, nodeUuid, rootNetworkUuid, substationsIds));
     }
 
-    @PutMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/network-modification/{modificationUuid}")
-    @Operation(summary = "Move a modification within or between containers (groups or composites)")
-    @ApiResponses(value = {@ApiResponse(responseCode = "200", description = "The modification order has been updated")})
-    public ResponseEntity<Void> moveModification(
+    @PutMapping(value = "/studies/{studyUuid}/nodes/{nodeUuid}/network-modifications/move")
+    @Operation(summary = "Move modifications within or between containers, possibly from another node of the same study")
```
Meklo: Expliciter qu'on ne peut faire des move que depuis un seul noeud par batch

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -1845,28 +1839,24 @@ public void moveNetworkModifications(
...
+            List<UUID> allModificationUuids = modificationInfos.stream().map(ModificationMoveInfos::modificationUuid).toList();
+            List<ModificationReference> allReferencesToMove = networkModificationService.getModificationReferences(allModificationUuids);
+            Map<UUID, List<ModificationReference>> referencesByModification = allReferencesToMove.stream()
+                    .collect(Collectors.groupingBy(ModificationReference::modificationUuid));
+            for (ModificationMoveInfos move : modificationInfos) {
+                UUID originNodeUuid = originNodeBySource.get(move.source());
+                boolean isSameNode = originNodeUuid == null || originNodeUuid.equals(targetNodeUuid);
+                moveElementReferences(move.source(), move.target(),
+                        referencesByModification.getOrDefault(move.modificationUuid(), List.of()),
+                        userId, studyUuid, targetNodeUuid, isSameNode);
```
Meklo: Move this logic to modification server

src/main/java/org/gridsuite/study/server/dto/modification/ModificationMoveInfos.java
Meklo: I renamed it insertBeforeUuid as suggested

src/main/java/org/gridsuite/study/server/controller/StudyController.java
```diff
@@ -651,46 +653,41 @@ public ResponseEntity<String> getAllMapData(
...
-            @RequestBody MoveModificationInfos moveModificationInfos,
+            @Parameter(description = "Node the modifications come from, defaults to the target node") @RequestParam(value = "originNodeUuid", required = false) UUID originNodeUuid,
+            @RequestBody List<ModificationMoveInfos> modificationMoveInfos,
             @RequestHeader(HEADER_USER_ID) String userId) {
+        UUID sourceNodeUuid = Objects.requireNonNullElse(originNodeUuid, nodeUuid);
+        studyService.assertIsStudyAndNodeExist(studyUuid, nodeUuid);
         studyService.assertIsNodeNotReadOnly(nodeUuid);
-        rebuildNodeService.moveNetworkModification(studyUuid, nodeUuid, modificationUuid, moveModificationInfos, userId);
+        if (!sourceNodeUuid.equals(nodeUuid)) {
+            // we don't cut - paste modifications from different studies
```
SlimaneAmar: Just `assertIsNodeExist` ?
Or not use `catch` !

src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
```diff
@@ -471,35 +469,28 @@ public void stopBuild(@NonNull UUID nodeUuid, @NonNull UUID rootNetworkUuid) {
...
+                .queryParam("originGroupUuid", originGroupUuid)
                 .queryParam("build", buildTargetNode);
-        if (beforeUuid != null) {
-            path.queryParam("before", beforeUuid);
-        }
 
         HttpHeaders headers = new HttpHeaders();
         headers.setContentType(MediaType.APPLICATION_JSON);
-        HttpEntity<Pair<List<UUID>, List<ModificationApplicationContext>>> httpEntity = new HttpEntity<>(body, headers);
+        HttpEntity<Pair<List<ModificationMoveInfos>, List<ModificationApplicationContext>>> httpEntity =
```
SlimaneAmar: Perhaps a mixed DTO with context ?

SlimaneAmar (approved):

src/main/java/org/gridsuite/study/server/controller/StudyController.java
Meklo: It previously threw a 403, so I just overloaded assertIsNodeExist to pass custom error codes otherwise some test breaks

src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
Meklo: IMO that would be a follow up PR, all endpoints directed at the network modification server (copy, duplication, composite insertion...) have this model `Pair<DTO, List<ModificationApplicationContext>>`. So if we change this pattern it should ideally done in one pass

PR #1099, opened 2026-09-16
Delete the leftover modifications to exclude with their owner

flomillot:
Remaining from https://github.com/gridsuite/study-server/commit/6d025b223e2f3135871cb2b6b3b9097998698797#diff-febcacd5a6da473c8d8d5231c272f4ec56c705a812e990eac86c26d6a689da86

The applicability per root network tag replaced the modifications a root network used to leave out, and their collection was unmapped without its table being dropped. The rows of the studies created before then still point at their root network node info, and the foreign key now refuses to delete it.

Any existing database is affected, and deleting a root network, a node or a study fails on it.

This maps the collection again so that hibernate deletes those rows with their owner. Nothing reads it, and the table is kept rather than dropped, in case the applicabilities ever have to be rebuilt from it: it will go in a later release.


src/main/java/org/gridsuite/study/server/networkmodificationtree/entities/RootNetworkNodeInfoEntity.java
```diff
@@ -113,6 +115,16 @@ public class RootNetworkNodeInfoEntity {
     })
     private NodeBuildStatusEmbeddable nodeBuildStatus;
 
+    // The modifications a root network used to leave out, replaced by the applicability per root network tag. Nothing
+    // reads this collection any more: it is only mapped so that the rows left in the table are deleted with their root
+    // network node info, which their foreign key would otherwise hold back. The table itself goes in a later release.
+    @ElementCollection
```
etiennehomer, line 121: Add something like : 
//TODO: to remove and generate changeset at release v2.42.0

etiennehomer (changes requested):
.

flomillot:
Superseded by #1100, which drops the foreign key instead of mapping the collection again: same fix without a mapping nothing reads, and without a delete per root network node info. The table and its rows are kept in both, until a later release drops them.

PR #1104, opened 2026-09-18
Require the write permission on shared modifications to rename a root network tag

flomillot:
Renaming a root network tag rewrites the applicabilities of the shared modifications of the study, which other studies use too: it now requires the write permission on each of them.

Adds `GET /studies/{studyUuid}/network-modifications/references/exists`, telling whether a study holds shared modifications, for the front to warn before a rename.

Goes with gridsuite/network-modification-server#901 and gridsuite/gridstudy-app#4209.


src/main/java/org/gridsuite/study/server/service/NetworkModificationService.java
```diff
@@ -724,6 +724,24 @@ public boolean hasModificationReferences(List<UUID> containerUuids) {
         return Boolean.TRUE.equals(restTemplate.getForObject(getNetworkModificationServerURI(false) + path, Boolean.class));
     }
 
+    /**
+     * @return the shared modifications the given containers point to, directly or through other shared modifications
+     */
+    public Set<UUID> getReferencedModifications(List<UUID> containerUuids) {
```
ghazwarhili:
```suggestion
    public Set<UUID> getReferencedModificationUuids(List<UUID> containerUuids) {
```

src/main/java/org/gridsuite/study/server/service/StudyService.java
```diff
@@ -415,6 +431,11 @@ private List<UUID> getStudyModificationGroupUuids(UUID studyUuid) {
     @Transactional
     public void updateRootNetworkRequest(UUID studyUuid, RootNetworkInfos rootNetworkInfos, String userId) {
         rootNetworkService.assertCanModifyRootNetwork(studyUuid, rootNetworkInfos.getId(), rootNetworkInfos.getName(), rootNetworkInfos.getTag());
+        String newTag = rootNetworkInfos.getTag();
+        if (newTag != null && !newTag.equals(rootNetworkService.getRootNetworkTag(rootNetworkInfos.getId()))) {
+            // If we rename the tag we check the permission.
```
ghazwarhili: // A rename requires WRITE permission on the shared modifications.
it is better

ghazwarhili:
I noticed that renaming a tag computes the shared modification tree twice:  assertCanRenameRootNetworkTag (study-server) calls "getReferencedModifications" sending request to network-modification-server which runs "getSharedModificationTrees" then, when saving "renameRootNetworkTag" makes a second request and recomputes the tree internally, @etiennehomer it's worth the renaming tag becomes slow for studies with deeply nested shared modifications ?

src/main/java/org/gridsuite/study/server/service/StudyService.java
flomillot: Ok

